PR #115, opened 2023-03-27
Semantics: initialization 

geo2a:
This PR enables a preparatory stage for the execution semantics: definition initialization phase.

In the initialization phase comprises the following activities:

* Travers the parsed Mir AST and construct the initial runtime configuration. More specifically:
  - create cells that hold functions together with their metadata: local declarations and basic blocks;
  - resolve any ambiguities encountered while parsing the basic block bodies: some of the Mir rvalues are highly ambiguous;
* Fail with a somewhat meaningful error message if the Mir source is invalid, i.e. contains duplicate locals or basic blocks.

Besides the work on the initialization semantics, this PR contains a significant refactoring of the syntax definition:
* the syntax definition is split up in several modules that are placed in several literate Markdown files;
* I've attempted to make some of the sorts more precise, which has caused a regression in the `rustc`-generated parsing tests. The tests that are failing now were very likely false successes, i.e. the parsed AST would not make much sense for the execution semantics.

It is possible that the current syntax definition does not make 100% sense, i.e. it is likely that we may need to refine it going further.


yanliu18 (commented):
@geo2a This looks nice, and very well organised. I was thinking of breaking down the big chunk of the MIR syntax file and grouping them into elements of the problem. (No offence to @virgil-serbanuta, your work is excellent!) Thank you!

yanliu18 (approved):

PR #176, opened 2023-07-12
Refactor/builtin identifiers

yanliu18:
This PR focused on refactoring the parsing rules for operations in RValue.
More tests are failed in this case, since the `Identifier (Operand, Operand)` pattern is less restrictive than explicit built-ins.

We will continue to refactor the parsing rules aiming to be consistent with [the rustc defined MIR syntax](https://github.com/rust-lang/rust/blob/master/compiler/rustc_middle/src/mir/syntax.rs).


geo2a:
@yanliu18 I'm approving this PR to unblock you, the changes look OK to me. I remember there was a way to update the integration tests' failure lists. You need to do that to not run the failing tests in CI.

geo2a (approved):

yanliu18:
@geo2a thanks for approving. I did run the `make compiletest-parse-fail` to update the failure list. BTW, these commands are really handy. I remeber you composed them. Big thank you!

I believe there are other issues with CI, there should be 2550 (now) test cases instead of 3197 (before submodule update). The number tells the submodule needs to be updated in CI.

PR #183, opened 2023-07-20
Bug reports `--bug-reports`

dkcumming:
Bug reports available for `kmir exec` through `--bug-report`.

### DIFF
--- a/kmir/src/kmir/__main__.py
+++ b/kmir/src/kmir/__main__.py
@@ -1,9 +1,11 @@
 from argparse import ArgumentParser
 from typing import Any
+from pathlib import Path
 
 from pyk.cli.utils import dir_path, file_path
 from pyk.ktool.kprint import KAstInput, KAstOutput
 from pyk.ktool.krun import KRunOutput
+from pyk.utils import BugReport
 
 from .kmir import KMIR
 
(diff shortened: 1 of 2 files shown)


Baltoli (approved):
Looks good to me if CI passes; the code checks currently need some changes I think

kmir/src/kmir/__main__.py
```diff
@@ -41,12 +43,13 @@ def exec_run(
...
     definition_dir: str,
     output: str = 'none',
+    bug_report: bool = False,
     ignore_return_code: bool = False,
     **kwargs: Any,
 ) -> None:
     krun_output = KRunOutput[output.upper()]
-
-    kmir = KMIR(definition_dir, definition_dir)
+    br = BugReport(Path(input_file).with_suffix('.bug_report')) if bug_report else None
```
dkcumming, line 51: @Baltoli I looked at the `KEVM` implementation for guidance, and the bug reports are being produced. But they are always produced at `{filename}.tar` not `{filename}.bug-report.tar` which I thought this line would of done. Before I figure out why, I just want to double check that `.bug_report` is preferred to included.

dkcumming: Awesome, thank you very much!

PR #186, opened 2023-08-02
Added `--max-depth` flag to `exec_run`

dkcumming:
Works as expected except that process ends in error if depth is less than the full process

### DIFF
--- a/kmir/src/kmir/kmir.py
+++ b/kmir/src/kmir/kmir.py
@@ -100,6 +100,7 @@ class KMIR:
         self,
         program_file: Union[str, Path],
         *,
+        depth: int | None = None,
         output: KRunOutput = KRunOutput.NONE,
         check: bool = True,
         temp_file: Optional[Union[str, Path]] = None,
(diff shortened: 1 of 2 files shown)


kmir/src/kmir/kmir.py
```diff
@@ -113,6 +114,7 @@ def run(program_file: Path) -> CompletedProcess:
                 pipe_stderr=True,
                 pmap={'PGM': str(self.mir_parser)},
                 bug_report=self.bug_report,
+                depth=depth,
```
dkcumming, line 117: @Baltoli Do you know why a call to `_krun` with a depth that interrupts the process would end in
```
[Error] krun: 
/home/daniel/.kbuild/mir-semantics/5a0ac5c/target/v6.0.24/llvm/interpreter 
/tmp/.krun-2023-08-03-09-04-52-Z4PGRkFugW/tmp.in.tUEaHwl7nA 2 
/tmp/.krun-2023-08-03-09-04-52-Z4PGRkFugW/result.kore

Command krun exited with code 4 for: /tmp/tmpkao_an5u
```

The `configuration` at the correct depth is printed beforehand as expected, but `_krun` then errors. This can be reproduced from `mir-semantics/kmir` with 
```
poetry run kmir run --definition-dir $(poetry run kbuild which llvm) --max-depth=2 --output pretty src/tests/integration/test-data/handwritten-rust/test-binop.mir
```

PR #195, opened 2023-08-15
Default definition paths

dkcumming:
Automatically connect to haskell and llvm backends after `make build`. Can be manually overridden with a different definition if desired.


kmir/Makefile
```diff
@@ -116,6 +116,7 @@ check-black: poetry-install
 # Kompilation
 
 kbuild-%:
-	$(POETRY_RUN) kbuild kompile $*
+	$(POETRY_RUN) kbuild kompile $* > $*.def; cat $*.def
```
dkcumming, line 119: I'm not sure if this is the best way to do this. I was having trouble with environment variables. `cat $*.def` is only there so when building it still prints the location of the definition, but this isn't strictly necessary.

kmir/src/kmir/__main__.py
```diff
@@ -50,7 +50,7 @@ def exec_run(
 ) -> None:
     krun_output = KRunOutput[output.upper()]
     br = BugReport(Path(input_file).with_suffix('.bug_report.tar')) if bug_report else None
-    kmir = KMIR(definition_dir, definition_dir, bug_report=br)
+    kmir = KMIR(definition_dir, None, bug_report=br)
```
dkcumming, line 53: Because I am passing `None`, both backends do need to be built. Maybe this is not preferable.

yanliu18:
Hi Daniel, thanks for the PR. It is a friendly and desirable feature.
I've tested locally, and there are two errors.
In summarise, there are three ways to call kmir:
1. ```poetry run kmir parse <parth to file.mir>```
2. In a shell,
```
    poetry shell 
    kmir parse <parth to file.mir>
```
3. Global install, 
```
     pip install dist/.whl 
     kmir parse <parth to file.mir>
```

This feature succeeded in 1 and failed in 2 and 3.

Testing in 2 failed with an error `FileNotFound (llvm.def)`, and testing in 3 failed with a type error

yanliu18:
> Hi Daniel, thanks for the PR. It is a friendly and desirable feature. I've tested locally, and there are two errors. In summarise, there are three ways to call kmir:
> 
> 1. `poetry run kmir parse <parth to file.mir>`
> 2. In a shell,
> 
> ```
>     poetry shell 
>     kmir parse <parth to file.mir>
> ```
> 
>  3. Global install,
> 
> ```
>      pip install dist/.whl 
>      kmir parse <parth to file.mir>
> ```
> 
> This feature succeeded in 1 and failed in 2 and 3.
> 
> Testing in 2 failed with an error `FileNotFound (llvm.def)`, and testing in 3 failed with a type error

2 should succeed. But if you test it under other directories, for example 
```
poetry shell
cd src/tests/integration/test-data/handwritten-mir/parsing
kmir parse statement-list.mir
```
It will fail with the file not found error.

dkcumming:
2. Should be solved

yanliu18:
@dkcumming Thanks. This is a handy feature and is helpful for development. However, I am not familiar with `pyk`. I'd like to invite Georgy to review.

Hi @ geo2a, can you take a look at this issue and advise if it is the best way of doing it?

dkcumming:
@geo2a I looked at `avm-semantics` and I am confused, because I cannot get the env vars to export properly with any configuration. Some help would be appreciated!

geo2a:
I think the approach I've been taking in KAVM is:
* build the semantics
* source the env vars manually, but with a shorthand, basically what you have in kmir/set_env.sh.

geo2a:
Actually, you can use [direnv](https://direnv.net/) to automatically set environment variables when you `cd` into the project directory. Here's the template [envrc](https://github.com/runtimeverification/avm-semantics/blob/master/envrc) for KAVM.

geo2a (approved):

PR #212, opened 2023-09-19
Removed custom `Layout` for `DefaultLayout`

dkcumming:
The default layout does not appear to be a problem with parsing.


kmir/k-src/mir-syntax.md
```diff
@@ -161,15 +161,4 @@ These constructs need to be disambiguated at runtime. See the `MIR-AMBIGUITIES`
...
-```k
-module MIR-PARSER-SYNTAX
-  imports MIR-SYNTAX
-
-  // Declaring regular expressions of sort `#Layout` infroms the K lexer to drop these tokens.
-  syntax #Layout  ::= r"(\\/\\/[^\\n\\r]*)" // single-line comments
-                    | r"([\\ \\n\\r\\t])"   // whitespace
-
-endmodule
-```
```
dkcumming, line 175: @geo2a @radumereuta in trying to discover why parser is including whitespace on some binary productions such as `+`, I tried using the default layout. Is there any benefit to using this layout that I have removed over the default layout? Otherwise it seems an appropriate simplification

geo2a: If default `#Layout` works, then I agree that we should use it.

geo2a (approved):

PR #213, opened 2023-09-26
Refactor/syntax update

dkcumming:
An update to the syntax to move closer to the `rustc` representation


kmir/k-src/mir-assert.md
```diff
@@ -0,0 +1,38 @@
...
+                      | DivisionByZero
+                      | RemainderByZero
+                      | ResumedAfterReturn
+                      | ResumedAfterPanic
+                      | MisalignedPointerDereference
+
+  syntax BoundsCheck ::= "\"index out of bounds: the length is {} but the index is {}\"" "," Operand "," Operand //Should this Oprand always uSize
+  syntax OverflowNeg ::= "\"attempt to negate " Operand ", which would overflow\""
+  syntax DivisionByZero ::= "\"attempt to divide `{}` by zero\"" "," Operand
+  syntax RemainderByZero ::= "\"attempt to calculate the remainder of " Operand " with a divisor of zero\""
```
virgil-serbanuta, line 21: Here and also below - it probably does not matter, but you're using spaces at the end of some tokens. Is it because you want to make sure that the last word ends at that point?

kmir/k-src/mir-lexer-syntax.md
```diff
@@ -20,7 +20,7 @@ This module defined the necessary `token` productions.
 ```k
   syntax IdentifierToken ::= r"[_a-zA-Z][_a-zA-Z0-9]*" [token]
   syntax LocalToken      ::= r"_[0-9]+"  [token]
-  syntax BBToken         ::= r"bb[0-9]+" [token]
+  syntax BBId         ::= r"bb[0-9]+" [token]
```
virgil-serbanuta, line 23: I think that this overlaps with `IdentifierToken`. If that's the case, then it may happen that BBIds get parsed as IdentifierTokens, so you should probably use `prec` to give a higher priority to `BBId`. I think that the same applies to `LocalToken`.

kmir/k-src/mir-lexer-syntax.md
```diff
@@ -39,7 +39,7 @@ Simplified forms of the [Rust literals](https://doc.rust-lang.org/reference/toke
 
 ```k
   syntax FilePosition ::= FileLineColumnStartLiteral LineColumnEndLiteral
-  syntax FileLineColumnStartLiteral ::= r"[^@ ]+:[0-9]+:[0-9]+:"  [token]
+  syntax FileLineColumnStartLiteral ::= r"[a-zA-Z0-9_/\\-\\.]*\\.rs:[0-9]+:[0-9]+:"  [token] // Adding slash (type-param-constraunts.mir LINE 42), and forcing '.rs' extension
```
virgil-serbanuta, line 42: s/constraunts/constraints/ , perhaps?

kmir/k-src/mir-types.md
```diff
@@ -284,6 +285,10 @@ module MIR-TYPE-SYNTAX
                               | ByteStringLiteral
                               | Int
                               | Bool
+                              // | CapitalIdent // TEMPORARY
+                              // | Identifier // IDENTIFIER IS INCORRECT AS FIRST POSITION CANNOT BE UPPER CASE. NEED TO PRINT ACTUAL REPRESETATION
```
virgil-serbanuta, line 289: s/REPRESETATION/REPRESENTATION/

I'm not sure exactly what you need here, but you could split `IdentifierToken`:
```
syntax FirstUpperIdentifierToken ::= r"[A-Z][_a-zA-Z0-9]*" 
syntax FirstOtherIdentifierToken ::= r"[_a-z][_a-zA-Z0-9]*"
```
then you could probably use something like this:
```
| FirstUpperIdentifierToken  [prefer]
| Identifier
```

virgil-serbanuta (commented):
Looks good to me. I think it would be better if someone more familiar with the current state of the project approves, but let me know if I should also do that.

geo2a (approved):

kmir/k-src/mir-assert.md
dkcumming: No I think that is just a mistake. I removed it

kmir/k-src/mir-lexer-syntax.md
dkcumming: That's actually an interesting point. I think I will investigate how rustc handles that conflict and write a test for it. For now I will add `[prefer]` to `BBId`

PR #223, opened 2023-10-10
Update K tag from Pyk

Baltoli:


.github/workflows/update-version.yml
```diff
@@ -30,6 +30,12 @@ jobs:
           poetry update
           git add pyproject.toml poetry.lock
           git commit -m 'kmir/{pyproject.toml,poetry.lock}: sync Poetry files' || true
+      - name: 'Update K release tag'
+        run: |
+          PATH="/home/github-runner/.local/bin:$PATH"
```
tothtamas28, line 35: Why is setting `PATH` necessary here?

tothtamas28 (approved):

.github/workflows/update-version.yml
Baltoli: Ah, I missed `echo ${HOME}/.local/bin >> $GITHUB_PATH` above. Will remove this line

PR #225, opened 2023-10-11
`clean` before `build`

dkcumming:
I ran into some issues extending the syntax and semantics where running `make build` carried through errors from the previous build. Removing `kdist/` and then `build` solved the problem. Unfortunately I don't have a reproducer for this.

### DIFF
--- a/kmir/Makefile
+++ b/kmir/Makefile
@@ -19,7 +19,7 @@ clean:
 	find . -type d -name __pycache__ -prune -exec rm -rf {} \;
 
 .PHONY: build-backends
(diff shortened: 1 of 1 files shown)


dkcumming:
@geo2a Is this expected behaviour? Or is this new and I should let someone know about the problem? Perhaps @tothtamas28 is interested by this.

geo2a (approved):
Approving this as a workaround, but it may indeed by a bug in `kbuild`/`kompile` as Tamas suggests.

dkcumming:
I couldn't get it to happen again sorry. If it does happen again ever I will make an issue.

PR #229, opened 2023-10-16
Check correct k version is used for `Make` `build-backends`

dkcumming:
This establishes a check that the correct version of `k` is installed, with instruction for how to install the correct version through `kup` if the wrong version is installed. This should protect developers from accidentally using the incorrect version, however it will also stop them from _intentionally_ using the incorrect version unless they change the script - an acceptable tradeoff.

### DIFF
--- a/kmir/Makefile
+++ b/kmir/Makefile
@@ -132,3 +132,13 @@ check-k-version:
 
 .PHONY: kbuild-*
 
+# Nix (doesn't perform version check)
+
+.PHONY: nix-build-backends
+
+nix-build-backends: clean nix-kbuild-llvm nix-kbuild-haskell
+
+nix-kbuild-%:
+	$(POETRY_RUN) kbuild kompile $*
+
+.PHONY: nix-kbuild-*
--- a/kmir/Makefile
+++ b/kmir/Makefile
@@ -131,14 +131,3 @@ check-k-version:
 	./check_k_version.sh
 
 .PHONY: kbuild-*
-
-# Nix (doesn't perform version check)
-
-.PHONY: nix-build-backends
-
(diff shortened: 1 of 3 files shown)


kmir/Makefile
```diff
@@ -121,11 +121,14 @@ check-black: poetry-install
 
 # Kompilation
 
-kbuild-%:
+kbuild-%: check-k-version
 	$(POETRY_RUN) kbuild kompile $*
```
dkcumming, line 125: Having this check here doesn't work for `nix`.

kmir/Makefile
```diff
@@ -121,11 +121,24 @@ check-black: poetry-install
...
+# Nix (doesn't perform version check)
+
+.PHONY: nix-build-backends
+
+nix-build-backends: clean nix-kbuild-llvm nix-kbuild-haskell
+
+nix-kbuild-%:
+	$(POETRY_RUN) kbuild kompile $*
+
+.PHONY: nix-kbuild-*
```
Baltoli, line 144: I'm happy with these changes in principle but I'd like to understand _why_ the Nix version check doesn't work; that might well be a bug we should fix in upstream K.

Baltoli: It could in theory be a different version, yes. `kup` provides the `--override PACKAGE VERSION` flag to change the version of a flake input in development. This might be something you want to allow, as a sort of moral equivalent to the user changing the script in development to get a different K version, or it might be something you want to forbid.

I'm happy either way; whatever option you prefer for your development and usage ergonomics is fine, but I just wanted to check that there wasn't a _bug_ in the Nix version of K that meant you couldn't get the version properly.

Baltoli (approved):

PR #241, opened 2023-10-29
Semantics/min max

dkcumming:
Core constants can that refer to the minimum and maximum of [certain types](https://github.com/rust-lang/rust/blob/347452e7e378d93e7ba29af05f022a5445e1f459/library/core/src/num/mod.rs#L1148), [example](https://github.com/runtimeverification/mir-semantics-compiletest/blob/master/ui/big-literals.mir#L217). This PR includes syntax and semantics for minimum and maximum constants for u8 - u128, and i8 - i128. However usize and isize are omitted currently as that depends on the architecture that the program is being run on, which we do not consider yet. There is also a test, albeit a strange one.

Unfortunately, these constants are not produced by the same looking rust function, for example x:u32 = u32::MAX in a rust file will compile to _1 = const _, however x:u32 = 4294967295 will compile to _1 = u32::MAX. There may be other compilation flags to help with this other than what we are [currently using](https://github.com/runtimeverification/mir-semantics-compiletest/blob/84ba5c279343b1c31c436538b902348700c62415/Makefile#L9).


geo2a (approved):

kmir/k-src/mir-rvalue.md
```diff
@@ -229,9 +229,40 @@ Evaluate a syntactic `RValue` into a semantics `RValueResult`. Inspired by [eval
...
+  rule minInt(i8)   => -128
+  rule minInt(i16)  => -32768
+  rule minInt(i32)  => -2147483648
+  rule minInt(i64)  => -9223372036854775808
+  rule minInt(i128) => -170141183460469231731687303715884105728
+```
+
 ### `Local` evaluation
 
 Locals only makes sense withing a function-like, hence we evaluate them as a contextual function that grabs the value from the function-like's environment:
```
Baltoli, line 268:
```suggestion
Locals only makes sense within a function-like, hence we evaluate them as a contextual function that grabs the value from the function-like's environment:
```

kmir/k-src/mir-rvalue.md
```diff
@@ -229,9 +229,40 @@ Evaluate a syntactic `RValue` into a semantics `RValueResult`. Inspired by [eval
   rule evalConstantValue(VALUE:StringLiteral)   => StringLitertal2String(VALUE)
   rule evalConstantValue(( ))                   => Unit
   rule evalConstantValue(VALUE:Bool)            => VALUE
+  rule evalConstantValue(VALUE:ConstEnumConstructor)  => evalPrimitiveBound(VALUE)
 //  rule evalConstantValue(_VALUE)              => "Error: evalConstantValue --- unsupported ConstantValue" [owise]
 ```
 
+#### Primitive type bounds TODO: usize depends on architecture, which we currently do not handle
```
Baltoli, line 236: The changes here all seem fine to me; let me know if you would like any pointers or advice from our experience with the C Semantics when you come around to implementing the architecture-specific bits.

Baltoli (approved):

PR #243, opened 2023-10-30
Semantics/recursion

dkcumming:
Added naive recursion. Does not support simultaneous recursion (e.g. `fib(n-1) + fib(n-2)`. Only looks to see if the function call is the same name as the containing function, and then checks on the top of the call stack for the depth. It then duplicates the stack frame as is, this currently carries over all the data from the original frame and does not clear it - this generally shouldn't matter as default values are probably atypically read, but all the same it is bad practice. There is a lot of unnecessary duplication in the copied stack frame, the basic blocks are the same and need not be duplicated at all. There are also general inefficiencies in the current build that ought be cleaned up.


geo2a (approved):
Looks good to me! We can have a different implementation is future if necessary.

kmir/src/tests/integration/proofs/simple-spec.k
```diff
@@ -84,20 +84,20 @@ module SIMPLE-SPEC
...
+    //             <value> _ => 42 </value>
+    //             ...
+    //           </localDecl>
+    //           <localDecl>
+    //             <index> 2 </index>
+    //             <value> 42 </value>
+    //             ...
+    //           </localDecl>
+    //           ...
+    //       </function>
```
dkcumming, line 102: Unsure why this proof is failing right now. There needs to be a revision of the entire proof runner. I think leaving this commented out is fair for now as it demonstrates that it was passing and a blame will show the commit it was commented.

PR #244, opened 2023-10-31
Added evaluation of tuples.

dkcumming:
This PR only concerns the evaluation of tuples directly, e.g. `_2 = (const 1_usize, const 2_i32) `. Accessing the fields is left to a different PR. The test that is added will not be able to pass until the fields are implemented.

### DIFF
--- a/kmir/k-src/mir-types.md
+++ b/kmir/k-src/mir-types.md
@@ -423,6 +423,7 @@ module MIR-VALUE
   imports BYTES
   imports MIR-TYPE-SYNTAX
   imports MIR-LEXER-SYNTAX
+  imports LIST
 
   syntax KItem ::= RValueResult
 ```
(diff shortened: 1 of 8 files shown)


kmir/k-src/mir-types.md
```diff
@@ -448,12 +448,13 @@ TODO: add more domain sorts
                     | "Unit"
                     | Bool
                     | "Never"
+                    | "(" MIRValueNeList ")"
                     | "UNIMPLEMENTED"
 
   syntax RValueResult ::= MIRValue
                         | MIRValueNeList
```
dkcumming, line 455: I am unsure if this should be an `RValueResult`, likely the `MIRValueNEList` is inside some other structure like a tuple or  a slice. The tuple is seen on line 451, and for a slice it would be `| "[" MIRValueNeList "]"`, so perhaps `MIRValue` is the only `RValueResult`

geo2a (approved):

PR #250, opened 2023-11-06
`Option` evaluation

dkcumming:
Basic evaluation for Standard Library / [core](https://github.com/rust-lang/rust/blob/master/library/core/src/option.rs) `Option` type. This does not offer full support for `Options`, requiring some more work on references to for full support. However this PR passes 6 more tests. Numerous failing tests have changed from `4` to `113`, a quick inspection of a couple of tests indicated that this was not from this PR as much as this PR allows execution of those tests to go further to then fail from something else.


kmir/k-src/mir-rvalue.md
```diff
@@ -364,6 +369,23 @@ Locals only makes sense within a function-like, hence we evaluate them as a cont
     requires PLACE_INDEX ==Int Local2Int(PLACE)
 ```
 
+### `Enum` evaluation
+```k
+  syntax MIRValue ::= evalEnumConstructor(FunctionLikeKey, EnumConstructor) [function]
+  //--------------------------------------------------------------
+  rule evalEnumConstructor(FN_KEY, Option :: < _TYPES > :: .ExpressionPathList :: Some ( OP , .OperandList ) ) => OptSome(evalOperand(FN_KEY, OP:Operand))
+    requires IdentifierToken2String(Option) ==String "Option" andBool IdentifierToken2String(Some) ==String "Some"
```
dkcumming, line 377: @Baltoli I couldn't find it in the documentation for K, but is there a way to be have uppercase sequences of characters not match as a variable. To be explicit `Option` above is a variable, but this rule is expecting that variable is actually only `Option` verbatim, which is added as a side condition. Wrapping in `""` creates a `String`. Can I signal that I want the upper case to be treated verbatim?

kmir/k-src/mir.md
```diff
@@ -515,6 +515,13 @@ or panics if the function-like or the block is missing:
         ...
        </k>
        <callStack> ListItem(FN_KEY) ... </callStack> [owise]
+
+  // Option Unwrap Call
+  rule <k> #executeTerminator(DEST_LOCAL:Local = Option :: < _TYPES > :: Unwrap :: .ExpressionPathList ( ARG , .OperandList ) -> ((NEXT:BBName _):BB)) 
+        => #executeStatement(DEST_LOCAL = #unwrap(ARG)) ~> #executeBasicBlock(FN_KEY, BBName2Int(NEXT)) ...
```
dkcumming, line 521: Treating the assignment as a statement might not be a proper way of doing things. It could be confused that the origin was from an actual `Statement`, however it came from a `Terminator`. I personally think this is the best way to do things anyway.

kmir/k-src/mir-rvalue.md
Baltoli: You want to declare `Option` as a token production of sort `IdentifierToken`; it will then parse as a token rather than as a variable. This is alluded to in Lesson 16 of the K tutorial, but not very clearly at all so I think we could probably improve the documentation there.

The issue is that the **rule parser** behaves differently to the program parser in a few ways to make it easier to write common-case rewrite rules. One of these is in the overlap between scanning sequences of characters; anything that can be scanned as a variable will be, under the assumption that users are more likely to write a rule mentioning variables than they are to mention an upper-cased literal token in some sort.

Here's a K definition demonstrating this difference:
```k
module UPPER-SYNTAX
  imports ID-SYNTAX

  syntax Foo ::= "foo" Id
               | "bar" Id
endmodule

module UPPER
  imports UPPER-SYNTAX

  syntax Id ::= "Option"  [token]
              | "Some"    [token]

  rule foo Option => bar Some [label(token)]
  rule foo A => bar A         [label(variable), owise]
endmodule
```

Baltoli (changes requested):
Some minor changes re. your `IdentifierToken` query; otherwise this looks fine to me!

kmir/k-src/mir-lexer-syntax.md
```diff
@@ -10,8 +10,12 @@ This module defined the necessary `token` productions.
                       | BBId
                       | DoubleHexDigitNoIntLetter
                       | Whitelisted
+  
+  syntax Whitelisted ::= "main" [token]
 
-  syntax Whitelisted ::= "transmute" | "unwind" | "count"
+  syntax Whitelisted ::= "transmute" [token] | "unwind" [token] | "count" [token]
+
+  syntax Whitelisted ::= "Option" [token] | "None" [token] | "Some" [token] | "unwrap" [token]
```
dkcumming, line 18: @Baltoli This `Whitelisted` production I think might have been intended for this. I have made the terminals tokens, I am unsure about exactly what happens if I do or don't add the `[token]` attribute. The tutorial seems to indicate it will be beneficial when constructing the AST. Does this result in simpler the KORE in the end?

kmir/k-src/mir.md
```diff
@@ -515,6 +515,13 @@ or panics if the function-like or the block is missing:
...
+       </k>
+       <callStack> ListItem(FN_KEY) ... </callStack>
+    requires IdentifierToken2String(Option) ==String "Option" andBool IdentifierToken2String(Unwrap) ==String "unwrap" [priority(48)]
```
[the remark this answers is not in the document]
dkcumming: I will do a review of these in the future, I believe there is a better way to organise things that should avoid this.

kmir/k-src/mir-lexer-syntax.md
Baltoli: This is a bit subtle and relies on some knowledge of KORE as the backends interpret it. The user manual [documentation](https://github.com/runtimeverification/k/blob/master/docs/user_manual.md#token-attribute) for `[token]` is a bit terse, but it gets to the key point which is that _a token sort is inhabited by domain values and cannot have any constructors declared_. You can think of a "domain value" as the backend's run-time (as opposed to parse-time) analogue of a terminal production in a grammar; they cannot have child nodes, and are more or less represented as a string with some sort information attached.

In KORE, a domain value is written as a special parametric constructor `\dv`, while in K it's `#token`. For example, when you write the boolean literal `true` in K, the backend initially generates the K term `#token("true", "Bool")`, then the KORE term `\dv{SortBool{}}("true")`.

There will be some small differences in the generated KORE depending on whether you make these sorts token sorts or not; I would guess that it's probably marginally better to be pattern-matching on token productions rather than constructors, but I'd need to do some combing through the generated code to confirm that.

In this specific case, I think the most idiomatic way to organise this is as follows:
1. Drop the `Whitelisted` sort entirely unless it's somehow mentioned in the MIR specification; when we see the token `Option` I presume we really do mean that it's an `IdentifierToken` rather than something else.
2. At the point at which you need to mention a specific identifier in a rule or configuration, introduce a token production for that identifier into the relevant semantics module.

For example, [here's](https://github.com/runtimeverification/c-semantics/blob/e3d8a3e812b3344f70719cb954fa4dedb89c4f80/core/semantics/language/exp/primary.k#L23-L43) an example from the C semantics where we write some rules mentioning specific identifiers. Doing it this way keeps the tokens as close to their point of use as possible, and avoids polluting these tokens into the global syntax module (i.e. your grammar as it stands means you can't use `Option` as a variable any more).

dkcumming: Thank you, this is a very good explanation.

kmir/k-src/mir-lexer-syntax.md
```diff
@@ -9,13 +9,17 @@ This module defined the necessary `token` productions.
                       | LocalToken
                       | BBId
                       | DoubleHexDigitNoIntLetter
-                      | Whitelisted
   
-  syntax Whitelisted ::= "main" [token]
+  syntax IdentifierToken ::= "main" [token]
```
Baltoli, line 13: It's worth also pushing these token productions down into the semantics modules that actually use them (i.e. declare the token for `main` alongside the rule / configuration that needs to mention `main`)

Baltoli (approved):

PR #255, opened 2023-11-13
Explicit `#halt` rule and `KMIRSemantics`

dkcumming:
Added `#halt` to the semantics to signify the end of execution. Created `KMIRSemantics` that inherits from [KCFGSemantics](https://github.com/runtimeverification/pyk/blob/f28b2f48f4512ca0ba0338fe584442f172f15222/src/pyk/kcfg/semantics.py#L11C15-L11C15) that we will define our our `terminal` and `cut-point` rules for. This PR will only define the `terminal` rule for `#halt`. 

As a side note, the smoke tests are removed as their purpose was served.


geo2a:
The problem was caused by the fact that the `kfcg_semantics` parameter of `legacy_explore` was instantiated with the class itself, rather than with an object. I've also added trivial implementation of the methods (otherwise the class remains abstract and cannot be instantiated).

kmir/k-src/mir.md
```diff
@@ -41,7 +41,7 @@ The `#return` rule is triggered by the `return` terminator. We need to give it d
 If we are, then we stop execution and enter the finalization phase. Otherwise, if we're not currently in `main`, we return control to the caller function.
 
 ```k
-  rule <k> #return(FUNCTION_KEY, Unit) => .K ... </k>
+  rule <k> #return(FUNCTION_KEY, Unit) => #halt ... </k>
```
Baltoli, line 44: This rule should be rewritten using the `token` productions idiom rather than with a side condition on the rule.

kmir/k-src/mir.md
```diff
@@ -50,9 +50,19 @@ If we are, then we stop execution and enter the finalization phase. Otherwise, i
...
     requires FUNCTION_KEY =/=K Fn(main)
+```
+
+The `#halt` construct is used to signify the end of execution. Any remaining items on the `<k>` cell will be removed.
+
+```k
+  syntax KItem ::= "#halt"
+  //----------------------
+  rule [halt]: <k> #halt ~> (_:MirSimulation => .K) ... </k>
+  rule         <k> #halt ~> (#initialized()  => .K) ... </k>
```
Baltoli, line 61: Is the intent of these rules to throw away the entire contents of the `<k>` cell following the `#halt`? If so you probably want to rewrite them without structural framing (`...`); I'm not 100% convinced they'll do what you expect here (but I might be wrong!).

kmir/k-src/mir.md
```diff
@@ -50,9 +50,19 @@ If we are, then we stop execution and enter the finalization phase. Otherwise, i
...
 endmodule
 ```
 
+
 `MIR-SYMBOLIC` is a stub module to be used with the Haskell backend in the future. It does not import `MIR-AMBIGUITIES`, since the `amb` productions seem to not be supported by the Haskell backend. We may need to consult the C semantics team when we start working on symbolic execution.
```
Baltoli, line 66:
```suggestion
`MIR-SYMBOLIC` is a stub module to be used with the Haskell backend in the future. It does not import `MIR-AMBIGUITIES`, since `amb` productions are not supported by the Haskell backend. We may need to consult the C semantics team when we start working on symbolic execution.
```

Baltoli: I can confirm this is the case!

kmir/src/kmir/kmir.py
```diff
@@ -11,17 +11,60 @@
...
+                return True
+            elif (
+                k_cell.arity == 2 and k_cell[0] == KMIR.halt() and type(k_cell[1]) is KVariable and k_cell[1].sort == K
+            ):
+                return True
+        return False
+
+    @staticmethod
+    def terminal_rules() -> list[str]:
+        terminal_rules = ['EVM.halt']
```
Baltoli, line 49: This looks suspicious to me; I presume it should be `MIR.halt` looking at the rest of the code?

Baltoli (changes requested):

kmir/k-src/mir.md
dkcumming: I actually had the same thought! I used EVM as a model, and saw [they used structural framing for this](https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md#exception-based). I figured that it must be working then, and I wasn't sure of the nuance that might be behind leaving it. I tested and it is working without it as far as I can tell, I think I will leave it off.

PR #257, opened 2023-11-15
Handwritten test expansion

dkcumming:
This PR will properly organise the `handwritten-rust/` directory to be easily navigable when trying to identify the coverage of semantic constructs. Different directories are used to distinguish between archetypes/features of the code. The tests in these directories aim to display reasonable coverage of the feature, that is to say, Rust that is typically used and not contrived edge cases. Edge cases and the boundaries of Rust are better handled by the [Rust tests/ui/ test suite](https://github.com/rust-lang/rust/tree/master/tests/ui), which we use a subset of in [this repository](https://github.com/runtimeverification/mir-semantics-compiletest) which is included in this project as a submodule.

Failing tests should be minimised to the smallest size that includes the failing feature.


yanliu18:
Hi @dkcumming, I feel these tests overlap with the compile-rs/ui test suits.
I am OK to approve it and use it for local testing of MIR semantics. 
Though, we might try to reuse the test suits as much as possible.

dkcumming:
All of them overlap with ui test suite, that's unavoidable. But often it is very hard to find a test that is short and testing just one particular thing in the ui test suite. Often many of them will rely on multiple things that we don't currently support. This list is meant to be a sort of unit test for particular features.

yanliu18 (approved):

PR #281, opened 2023-12-06
Included `#panic` and `#internalPanic` in terminal rules.

dkcumming:
This stops the verifier from potentially infinitely spinning on panics.


kmir/k-src/mir.md
```diff
@@ -711,9 +711,7 @@ These are internal panics that are specific to KMIR.
 These panics are not specific to KMIR and caused by program-level reasons, i.e. assertion violations.
 
 ```k
-  rule <k> #panic(_FN_KEY, _PANIC, _MSG) ~> (_ITEM:KItem => .K) ... </k>
-       <returncode> 4 => 2 </returncode>
-  rule <k> #panic(_FN_KEY, _PANIC, _MSG) ... </k>
```
dkcumming, line 716: I don't think we need both of these.

kmir/src/kmir/kmir.py
```diff
@@ -29,24 +29,27 @@ class KMIRSemantics(KCFGSemantics):
     def is_terminal(self, cterm: CTerm) -> bool:
```
Baltoli, line 29: How come the logic for `is_terminal` doesn't reference the `terminal_rules` function? Where is the latter used elsewhere in the code, if not here?

Baltoli (commented):
A couple of questions; will approve when resolved but I'm not sure they will need explicit changes to the code.

kmir/k-src/mir.md
```diff
@@ -711,9 +711,7 @@ These are internal panics that are specific to KMIR.
...
+  rule [panic]: <k> #panic(_FN_KEY, _PANIC, _MSG) ~> (_ITEM:KItem => .K) ... </k>
```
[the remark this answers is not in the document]
dkcumming: I'm not sure what the intention was exactly as this was existing before I joined the project. @geo2a, I can see this was several months ago, but do you recall? I can see that it is only going to throw a singleton away as the error code as `2` will meant the rule only fires once.

kmir/src/kmir/kmir.py
dkcumming: My understanding might not be correct, but my understanding is that the `terminal_rules` function is used to structure or stop `rpc` calls in some way. However, that may be the case for [evm-semantics](https://github.com/runtimeverification/evm-semantics/blob/6d16eee9a2ca66a141db29a6c8b24dc5fcf61431/kevm-pyk/src/kevm_pyk/kevm.py#L141) because they are referencing it elsewhere, and I am not there yet so it is not used at all.

Baltoli: I guess take a look and see if the logic in `is_terminal` that checks for labels that are terminal can be refactored to use the `terminal_rules` function; if it can't then no worries

PR #291, opened 2023-12-11
Kompile uses new type inference algorithm

dkcumming:
A new type inference algorithm has been introduced to K. This PR opts into using that algorithm by setting the appropriate flags in `kbuild.toml`


Baltoli (approved):
Looks good; when https://github.com/runtimeverification/pyk/pull/767 is finalised and merged you should be able to use the new feature directly via Pyk.

Baltoli (changes requested):
Ah, in fact it [doesn't look like this is in fact supported](https://github.com/runtimeverification/pyk/blob/master/src/pyk/kbuild/project.py#L109-L127) via `kbuild.toml`. Probably best to wait and use the Pyk feature I linked above.

yanliu18:
I am popping this up since the type-inference feature is supported by pyk by PR https://github.com/runtimeverification/pyk/pull/767.

dkcumming:
Created an issue [here](https://github.com/runtimeverification/pyk/issues/788)

dkcumming:
@Baltoli Things should be working now, I am still a bit unsure how to determine if it is using the new algorithm or not. How would be determine this? If you would like to see differences in timing I am recording local timing in #308

Baltoli (approved):
Looks good to me! Timing probably won't be _hugely_ different I suspect.

PR #294, opened 2023-12-14
Create GLR parser at build time

dkcumming:
Instead of checking that the GLR parser is built when creating a KMIR object and building one if it is missing, it is now created anytime the LLVM definition is created via `make` and it is confirmed to exist when creating the KMIR object.


kmir/src/kmir/__main__.py
```diff
@@ -51,11 +51,13 @@ def exec_version(**kwargs: Any) -> None:
     print(f'KMIR Version: {VERSION}')
 
 
-def exec_init(llvm_dir: str, **kwargs: Any) -> KMIR:
+def exec_init(llvm_dir: str, **kwargs: Any) -> None:
     print(
-        'WARN: "init" was seen in args, this calls an internal function. If a file is named "init", it must be renamed'
+        'kmir init: Building GLR parser'
```
dkcumming, line 56: I think the warning is actually not necessary, and it shows some confusion on my part. Just a log to say we are building the parser is fine.

kmir/Makefile
```diff
@@ -142,6 +142,7 @@ check-black: poetry-install
 
 kbuild-%: check-k-version
 	$(POETRY_RUN) kbuild kompile $*
+	@./build_glr.sh $* $(POETRY_RUN)
```
dkcumming, line 145: This one might be a little confusing. Nix can't run poetry, so it when nix calls `make ...` it sets `POETRY_RUN=` so it is empty. This script will either be passed two args, the build target, and "poetry run" or nothing. And this will determine if it uses `poetry run` in the script.

kmir/build_glr.sh
```diff
@@ -0,0 +1,15 @@
...
+    case $# in
+
+    1)
+        kmir init $(kbuild which llvm)
+        ;;
+    *)
+        poetry run kmir init $(poetry run kbuild which llvm)
+        ;;
+    esac
+fi
```
dkcumming, line 15: I actually don't really want to use this script at all, but I couldn't get the conditional logic to work in the Makefile. Makefile stuff is not my strong suit.

yanliu18:
Here is the final result of testing on my machine:
- `gen-glr-bison-parser`  is the correct option in `kompile`, but it is not accepted by `kbuild`, due to [the reason](https://github.com/runtimeverification/pyk/issues/771) you've found;
- `gen-bison-parser = "true"`  works with kbuild, but failed on our semantics.  
 > nix building this bison parser still fails on my machine, refer to [this issue](https://github.com/runtimeverification/k/issues/3879).
 
The solution to this PR can be 
1. adapt our syntax definition to bison parser acceptables; 
2. request pyk to add gen-glr-bison-parser  to kbuild.py. 

I would prefer with option 2 @tothtamas28.
Meanwhile, I felt sorry that I must messed up the results yesterday since I was testing out different solutions. But I am glad that via the discussion today, I’ve finally located the magic kbuild.

yanliu18:
Thus, I'll pause the review until issue https://github.com/runtimeverification/pyk/issues/783 is addressed.

tothtamas28:
> request pyk to add gen-glr-bison-parser to kbuild.py

Opened a PR for it: https://github.com/runtimeverification/pyk/pull/785

dkcumming:
Blocked on #295

flake.nix
```diff
@@ -68,7 +68,7 @@
               export KBUILD_DIR=".kbuild"
               patchShebangs .
               make build-backends VERSION="${k_release}" POETRY_RUN=
-              kmir init $(kbuild which llvm)
+              kmir gen-glr-parser $(kbuild which llvm) --verbose
```
yanliu18, line 71: With Pyk PR#785 https://github.com/runtimeverification/pyk/pull/785 Merged, this line should be removed. The parser should be built at the `kbuild kompile llvm` step.

kmir/src/kmir/__main__.py
```diff
@@ -8,7 +8,7 @@
 from pyk.cterm import CTerm
 from pyk.kast.outer import KApply, KClaim, KRewrite
 from pyk.kcfg import KCFG
-from pyk.ktool.kprint import KAstInput, KAstOutput
+from pyk.ktool.kprint import KAstInput, KAstOutput, gen_glr_parser
```
yanliu18, line 11: Similarly, this change should be reverted too.

kmir/src/kmir/__main__.py
```diff
@@ -51,11 +51,11 @@ def exec_version(**kwargs: Any) -> None:
     print(f'KMIR Version: {VERSION}')
 
 
-def exec_init(llvm_dir: str, **kwargs: Any) -> KMIR:
-    print(
-        'WARN: "init" was seen in args, this calls an internal function. If a file is named "init", it must be renamed'
-    )
-    return KMIR(llvm_dir, llvm_dir)
+def exec_gen_glr_parser(llvm_dir: str, **kwargs: Any) -> None:
```
yanliu18, line 54: This funtion is not need too?

yanliu18 (changes requested):

kmir/src/kmir/__main__.py
dkcumming: Maybe not, but then we are going to need to redesign the test runner (not opposed to that occurring but not sure it is a goal for this PR), which I am not sure how big a job that is. Right now we build everything manually in python as [fixtures](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html).

kmir/src/tests/conftest.py
```diff
@@ -68,7 +69,9 @@ def project() -> Project:
 
 @pytest.fixture(scope='session')
 def llvm_dir(kbuild: KBuild, project: Project) -> Path:
-    return kbuild.kompile(project, 'llvm')
+    llvm_dir = kbuild.kompile(project, 'llvm')
+    gen_glr_parser(llvm_dir / 'parser_Mir_MIR-SYNTAX', definition_dir=llvm_dir, module='MIR-SYNTAX', sort='Mir')
+    return llvm_dir
```
dkcumming, line 74: @yanliu18 Here is where that function is used

dkcumming: Actually wait, I might be completely wrong. Let me check

kmir/src/kmir/__main__.py
dkcumming: Ignore this, I was wrong!

yanliu18 (approved):

PR #319, opened 2024-02-01
More proofs 2

dkcumming:
This PR continues the work of #305, which has the goal of completing [Milestone 7](https://github.com/runtimeverification/mir-semantics/milestone/7)
 - Small changes to allow `add.k` claim to prove.
 - Added new proof with assert statement.

Two problems that are now realised are:
1. Every token must be explicitly declared in `verify.k`, not just `LocalToken` that had conflict with preceding underscore of variables in k: `_1`. Addressed in #320
2. Claims require EVERY cell for `localDecls` and `basicBlocks` be listed with some minimal information. Addressed in #321


dkcumming:
@sskeirik, the initial plan for this PR is not entirely relevant now given the current proposal. All the same the proofs that are added are working and I think can't hurt to merge.

sskeirik (approved):

PR #322, opened 2024-02-01
Hotfix/add.k

dkcumming:
Fixed a couple of bugs that stopped `add.k` from being parsed by the prover. 
- Indices for `<localDecl>` are now `Local` instead of `Int`
- `num` in `verify.k` should of be `IdentifierToken` instead of `Identifier`


sskeirik (approved):
LGTM

Issue #320, opened 2024-02-01
Proofs require every token requires explict declaration in `verify.k`


dkcumming:
Perhaps an extension to an old problem that was never fully solved in #209. It is now clear that ALL tokens must be explicitly listed in the verification module in order for the the symbolic backend parser to successfully parse them [link](https://github.com/runtimeverification/mir-semantics/blob/d2ca4cee00d2490ff6aec3934ab67b09dd61452d/kmir/src/tests/integration/test-data/prove-rs/verify.k#L24-L27), like so:
```
    syntax LocalToken ::= "_0" [token]
    syntax LocalToken ::= "_1" [token]
    ...
    syntax BBId ::= "bb0" [token]
    syntax BBId ::= "bb1" [token]
    ...
    syntax IdentifierToken ::= "core" [token]
    syntax IdentifierToken ::= "panicking" [token]
    syntax IdentifierToken ::= "panic" [token]
    syntax IdentifierToken ::= "\"assertion failed: num == 42\"" [token]
``` 

This is not resolved by added regex such as:
```
syntax IdentifierToken ::= r"[_a-zA-Z][_a-zA-Z0-9]*" [prefer, token, prec(2)]
```
A type of approach that was previously recommended in #209. While listing out the tokens is an immediate work around, it is completely impractical for KMIR as the project scales to being able to handle larger programs.

dkcumming:
@Baltoli has recommended that we speak to the `Kontrol` team to see how they handle the issue. And also mentions that `syntax IdentifierToken ::= "\"assertion failed: num == 42\"" [token]` should not be encoded as an `IdentifierToken` but as a `String`

PR #323, opened 2024-02-05
Updates to proof feedback.

dkcumming:
- Replaced printing of information for passing and failing proofs in `exec_prove`. 
- Updated `prove` so pytest catches all failures in `prove.py`.
- Created test harness for `show-proof` that compares passing or failing proofs with expected output.
    - Also able to update the expected output by passing a flag to the test runner


kmir/src/tests/integration/test_prove.py
```diff
@@ -38,13 +39,45 @@ def test_handwritten(
...
             spec_file=spec_file,
             save_directory=use_directory,
             smt_timeout=300,
             smt_retry_limit=10,
         )
+
+        # Check if passed proofs are in show
+        if len(passed) != 0:
+            for proof in passed:
+                if proof.id in SHOW_TESTS:
```
yanliu18, line 53: Please create a seperate testing file for `show_proof` and testing the `show_proof` function instead of `prove` function.

kmir/src/tests/integration/utils.py
```diff
@@ -100,3 +100,9 @@
 PROVE_FAIL_FILE = TEST_DATA_DIR / 'handwritten-prove-fail.tsv'
 PROVE_EXCLUDE = ['verify.k']
 PROVE_FAIL = {test.split('\t')[0] for test in PROVE_FAIL_FILE.read_text().splitlines() + PROVE_EXCLUDE}
+
+##--------------------
+# show-proof test data
+##--------------------
+SHOW_FILE = TEST_DATA_DIR / 'show-proof'
```
yanliu18, line 107: It occurs to me that there is only one test for `show_proof`.
It is fine to have this single test.
Please create a seperate folder for `show_proof` tests related files.
SInce `verify.k` hardcoded the testcase (`num == 42`), it is fine to hardcode the expected result in the code.

kmir/src/tests/integration/test-data/show-proof
yanliu18, line 1: Is this file generated?
I am thinking if it can be removed.

yanliu18 (changes requested):

kmir/src/tests/integration/test-data/prove-rs/verify.k
```diff
@@ -18,6 +18,11 @@ module VERIFY
...
 
-    syntax Identifier ::= "num" [token]
-    // syntax Identifier ::= "temp" [token]
+    syntax IdentifierToken ::= "num" [token]
+    // syntax IdentifierToken ::= "temp" [token]
+
+    syntax IdentifierToken ::= "core" [token]
+    syntax IdentifierToken ::= "panicking" [token]
+    syntax IdentifierToken ::= "panic" [token]
+    syntax IdentifierToken ::= "\"assertion failed: num == 42\"" [token]
```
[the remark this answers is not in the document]
dkcumming: Every `IdentifierToken` that gets produced needs to be explicitly listed somewhere in a `.k` file right now. There will be a lot more coming as more proofs get written. We don't really want anyone doing proofs to have to list these out, and I have an issue for them in #320 that will need help from the K team again. I figured grouping them together made sense but they can be spread into the proof files. Either way we want them removed.

README.md
```diff
@@ -162,6 +162,7 @@ Use `make` to run common tasks (see the [Makefile](Makefile) for a complete list
 * `make format`: Format code
 * `make test-unit`: Run unit tests
 * `make test-integration`: Run integration tests
+* `make test-integration TEST_ARGS="--update-expected-output"`: Run integration tests, updating expected results for `show-proof` tests
```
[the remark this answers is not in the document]
dkcumming: I thought yes, this is what Everett was talking about in the meeting for the same harness that tests show can be used to update the expected output. This is the kontrol version he directed us towards ([link1](https://github.com/runtimeverification/kontrol/blob/0c089b7b5f073412b3919832d7cc2478da2eccd3/src/tests/integration/test_foundry_prove.py#L136), [link2](https://github.com/runtimeverification/kontrol/blob/0c089b7b5f073412b3919832d7cc2478da2eccd3/README.md?plain=1#L67))

kmir/src/tests/integration/test_prove.py
dkcumming: This is how I thought to implement it at first too, but I see why kontrol went the way of including the check in their [prove tests](https://github.com/runtimeverification/kontrol/blob/0c089b7b5f073412b3919832d7cc2478da2eccd3/src/tests/integration/test_foundry_prove.py#L132-L181). Show depends on a proof to be run to use for show, so if we separate it out I think it would end up with duplication of running proofs. This structure we just store what `proof.id` we have a saved show output for, and those proofs get checked for show as they are already being run.

kmir/src/tests/integration/utils.py
dkcumming: Well this is only just being integrated so there is one proof with the integration, but there will be more coming now this architecture is set up. Wouldn't it make sense to leave this so that it can be populated easily?

kmir/src/tests/integration/test-data/show-proof
dkcumming: No this file is the file I mentioned a few comments back where you list out what claims you want to be used for the `show-proof` tests. It's the equivalent of this file in [kontrol](https://github.com/runtimeverification/kontrol/blob/0c089b7b5f073412b3919832d7cc2478da2eccd3/src/tests/integration/test-data/foundry-show).

kmir/src/tests/integration/utils.py
ehildenb: Better to have a single test to start, and as we come up with more cases where proofs need to be demonstrated to show, then it's easy to add more. Always best to set-up a robust testing harness like this ahead of time so it's easier to add to it later if needed.

README.md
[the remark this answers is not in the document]
ehildenb: Yep, the idea is that you should be able to run just a single command and updated all the expected output.

ehildenb (approved):

kmir/src/tests/integration/test-data/show-proof
ehildenb: You can add test names to this file over time as we discover more cases where `show` is not behaving as we expect. Then we fix the bug, and add it to this file, then run `make test-integration TEST_ARGS='--update-expected-output'` to get it to update the files.

Issue #303, opened 2023-12-20
Add ability to specify a particular claim in a file to prove or view


dkcumming:
Proving and viewing failed proofs only works at the file level, the ability to select a proof from a file would be very handy.

yanliu18:
PR #312 should have provided this functionality in `show-proof` and `view-proof`.
Note that, the `claim label` should be the same as the `proof id` in the current design, could be invalid if pyk changes its design.

yanliu18:
@dkcumming would you verify if this issue is fixed?

dkcumming:
For `view-proof` and `show-proof` this functionality works, but we aren't able to optionally restrict the prover to only prove a particular claim in a spec file currently. The whole spec file is provided and it proves every claim inside.

yanliu18:
I see. We should be able to add a cmd option to do this by providing the claim label.
If it is not urgent, I can do it after I am back.

PR #325, opened 2024-02-09
Add feature: prove a single claim in `kmir prove`

yanliu18:
Implement the feature requested in issue #303 to prove a specified claim.

Change Log:
* Add `--claim` option to `kmir prove`, the argument to `--claim` is the targeted single claim to be proved.
  usage example: `kmir prove ../demo/simple-spec.k --claim statement.1`
                           or `kmir prove ../demo/simple-spec.k --claim SIMPLE-SPEC.statement.1`
* Add `--claim-list` option to `kmir prove`, when specified, KMIR will print a list of claim labels in the provided specification file.

Note that, although not stated in the original issue, there could be further requested features such as :
1. prove a list of claims
2. print a list of specs in a specific module.

The implementation left space for adding that feature.


kmir/src/kmir/kmir.py
```diff
@@ -110,8 +110,11 @@ def rpc_session(self, server: KoreServer, claim_id: str, trace_rewrites: bool =
                 )
 
     # A wrapper on KProve's get_claims
-    def get_all_claims(self, spec_file: Path) -> list[KClaim]:
-        return self.mir_prove.get_claims(spec_file)
+    def get_all_claims(self, spec_file: Path, claim_label: Optional[str]) -> list[KClaim]:
```
ehildenb, line 113:
```suggestion
    def get_all_claims(self, spec_file: Path, claim_label: str | None) -> list[KClaim]:
```

kmir/src/kmir/prove.py
```diff
@@ -1,8 +1,8 @@
 import logging
 from pathlib import Path
-from typing import Any, Final, Iterable
+from typing import Any, Final, Iterable, Optional
```
ehildenb, line 3: We use `_ | None` instead of `Optional[_]` throughout all our codebases.

kmir/src/kmir/kmir.py
ehildenb: Also, `get_claims` already supports a potentially empty list, so we probably don't need this function.

kmir/src/kmir/cli.py
```diff
@@ -102,6 +102,12 @@ def create_argument_parser() -> ArgumentParser:
         type=dir_path,
         help='Path to Haskell definition to use.',
     )
+    prove_subparser.add_argument(
+        '--claim',
+        default=None,
+        type=str,
+        help='Provide the claim label for proving a single claim',
+    )
```
ehildenb, line 110: I would make this an array, and allow the user to specify multiple claims, rather than just a single claim.

Also, we should not re-implement this option here. We should re-usie the upstream `KCLIOptions.spec_args`, and add to it: https://github.com/runtimeverification/pyk/blob/e1519523721e84eec783019f7c061121bc50241e/src/pyk/cli/args.py#L137

Here in KEVM, we re-use those arguments: https://github.com/runtimeverification/evm-semantics/blob/9e88278f72aebf74d94b928654b6364086bf165a/kevm-pyk/src/kevm_pyk/__main__.py#L736

So the idea is that pyk exposes some common functionality (`KProve.get_claims`), and then the generic options exposed via pyx's CLI module above matches that functionality directly. So the more we re-use the upstream functionality, the more things will "just work", in the sense that we don't need wrappers like `KMIR.get_all_claims` to handle the arguments.

kmir/src/kmir/prove.py
yanliu18: KMIR uses `Optional` throughout the code base. 
Is there any implementation issues that requires `_ | None` than `Optional`?

kmir/src/kmir/cli.py
yanliu18: I am not sure why `KCLIOptions.spec_args` is not used in the first place. 
The code above was inherited from previous ones.
@dkcumming do you have any idea of why?

But if you want that way, would you please submit a seperate issue regarding the changes wanted?
It seems a much bigger changes than this PR aims at.

kmir/src/kmir/kmir.py
yanliu18: `get_all_claims()` is a simple wrapper on pyk's implementation to make KMIR's code succinct.
Direct use of pyk's function would make the calling path too long. 

I'd prefer the current implementation.

kmir/src/kmir/cli.py
dkcumming: From memory, I think what happened is that since mir-semantics wasn't already using KCLIArgs I just worked with the framework we already had when introducing the prover code. In retrospect I should of made an Issue so we could of discussed introducing this, but I guess I didn't think of it and was focused on the prover.

yanliu18: @dkcumming I had the same impression. 
@ehildenb Well, I do not think I will make changes in this PR.
I'll suggest submit an issue in future to refactor the KCLIArgs. I would suggest maybe delay the work on it such that more requested feature could be icorperated.

ehildenb: Please make such an issue befor eresolving, and link it here. See https://runtimeverification.com/blog/github-guidelines-for-collaborative-coding

kmir/src/kmir/prove.py
ehildenb: We are trying to use consistent codestyle everywhere, so we should open a separate issue to make this codebase consistent with our codestyle elsewhere. @tothtamas28 I thought that `Optional[X]` vs `X | None` was supposed to be caught by our linters, is this repository not setup correctly with our python project template? https://github.com/runtimeverification/python-project-template

Please make an issue and link it here to get the codestyle updated.

tothtamas28: > is this repository not setup correctly with our python project template?

It is, it just needs to be updated (`cruft update`), and the `make pyupgrade` step added to CI.

kmir/src/kmir/cli.py
yanliu18: > Please make such an issue befor eresolving, and link it here. See https://runtimeverification.com/blog/github-guidelines-for-collaborative-coding

Sorry, what do you mean? I am not sure what is the expected result should. Would you please do it? 
I can link it when you made the issue. 
Thank you.

kmir/src/kmir/prove.py
yanliu18: @tothtamas28 We are not aware of these options. The project was created by you if I remember correctly.
But it will be great if there could be a documentation about what should be done to correctly set up the project. 
Do you have one? Or maybe you can write an updated one so far?

Meanwhile, would you please submit an issue in this repo regarding the desired settings?
This will be of great help to the new team @sskeirik.

yanliu18: Thank you, @tothtamas28. 
Now is it safe to resolve this conversation? @ehildenb

kmir/src/kmir/cli.py
ehildenb: An issue to refactor this repository to use upstream KCLIArgs class instead of making its own should be made, and then linked here.

yanliu18: > An issue to refactor this repository to use upstream KCLIArgs class instead of making its entirely own CLI argument handling should be made, and then linked here.

Done https://github.com/runtimeverification/mir-semantics/issues/330

kmir/src/kmir/__main__.py
```diff
@@ -129,9 +131,15 @@ def exec_prove(
     # We provide configuration of which backend to use in a KMIR object.
     # `use_booster` is by default True, where Booster Backend is always used unless turned off
 
+    if claim_list:
+        claim_labels = get_claim_labels(kmir, spec_file)
+        print(*claim_labels, sep='\n')
+        sys.exit(0)
```
ehildenb, line 137: Is it intended that this is left here? Or is this for testing?

yanliu18: Intended for pretty printing of the list of the claim labels.

ehildenb (approved):

PR #347, opened 2024-03-25
Fixup circular imports and typos

sskeirik:


kmir/k-src/mir-types.md
```diff
@@ -526,14 +524,13 @@ Hooked functions
...
-  imports MIR-SYNTAX
+  imports MIR-TYPE-SYNTAX
 ```
 
 We use several hooks which convert between token and string representations:
 
 ```k
-  syntax String ::= StringLiteral2Sring(StringLiteral) [function, total, hook(STRING.token2string)]
-  syntax StringLiteral ::= String2SringLiteral(String) [function, total, hook(STRING.string2token)]
+  syntax String ::= StringLiteral2String(StringLiteral) [function, total, hook(STRING.token2string)]
```
mariaKt, line 533: Why was `StringLiteral2Sring` removed?

kmir/k-src/mir-statement.md
```diff
@@ -1,5 +1,5 @@
 ```k
-requires "mir-rvalue.md"
+requires "mir-rvalue-syntax.md"
```
mariaKt, line 2: It looks like `requires "mir-place.md"` can also be removed (mir-rvalue-syntax.md requires mir-operand.md requires mir-pace.md).

mariaKt (approved):
Thank you for removing the circular imports and the cleanup.

As far as the `requires` for the modules, I generally appreciate them being explicit in each module for its dependencies rather than relying on an underlying module including them, because I think that inferring them (a) is more difficult to maintain and (b) may lead to unintended errors if a module changes. If this approach is better/more efficient, I am OK with merging this though. I have left a comment and a question.

kmir/k-src/mir-types.md
sskeirik: This was removed because I didn't see it being used; I suppose it is a bit out of scope.

kmir/k-src/mir-statement.md
sskeirik: Thanks for this. I will remove it. I am not 100% committed to having only required imports vs. all dependent imports---I can see it going both way.

For now, I'll commit this since I think the other PR changes are good. We can add back in these imports in a later version if we decide that going that route is better.

PR #349, opened 2024-04-10
Add options classes and use in functions

nwatson22:
Extend Options from pyk to centralize the location of default values for command line options. Adds option groups in `cli.py` and top-level command options groups in the foundry-related files.


sskeirik (approved):
So, I now have a basic understanding of the options and arguments classes that pyk provides and how they can be subclasses by semantics. This looks good. Thanks.

PR #353, opened 2024-04-17
Fix oversight in K dependency update

Baltoli:
Previously, the dependency update PR wasn't updating the Poetry project file at the same time as the other relevant parts of the project because it hadn't been manually changed initially. This PR addresses the oversight and brings the dependency back up to date.

It also fixes some code that needed updating to follow the most recent versions of Pyk.


flake.nix
```diff
@@ -26,6 +26,11 @@
           kmir-pyk = poetry2nix.mkPoetryApplication {
             python = nixpkgs-pyk.python310;
             projectDir = ./kmir;
+            src = rv-utils.lib.mkPykAppSrc {
+              pkgs = import nixpkgs { system = prev.system; };
+              src = ./kmir;
+              cleaner = poetry2nix.cleanPythonSources;
+            };
```
dkcumming, line 33: Does this just clean stale data that might be around?

dkcumming (approved):
Wow good catch with the `pyk` version not being updated! Everything LTGM.

flake.nix
Baltoli: Sort of; in order for Nix builds to be reproducible they usually define a cleaning step that removes files not relevant to the build. Here the reference to `cleanPythonSources` is a function that removes things like git ignored files, python compiled objects, etc.

Normally Poetry2Nix handles this for us transparently, but to work round an upstream bug we need to call a function in our central Nix tools repo that patches some of KMir's code to be compatible with Nix.

PR #365, opened 2024-06-14
Re-initialise project from scratch

Baltoli:
As discussed, this PR sets the project up with a fresh templated repository; the new semantics based on stable MIR is unlikely to share any meaningful implementation or tests with the old textual semantics, and so it's probably going to be easiest to begin with a fresh project rather than trying to rewrite incrementally.

I'm open to a less invasive approach if there's some element of the project setup that I haven't appreciated, but I wanted to open this PR to get the ball rolling on the new tooling.

The old version of the project is retained on the `textual-mir` branch, and it's easy to pull individual files across from the last state on that branch:
```console
$ git checkout origin/textual-mir $FILE
```

Versioning and CI are also retained so as to keep a clean release history for the project; I've bumped the version number to `0.3.0` as part of this change. Tests will be run when written, but are currently left at the "hello world" state deliberately.

The other deliberate change made here is to **not** cut releases until we have something more tangible built here; once the stable-MIR version of the project is more mature we can begin to do so.

Follow-up PRs:
- https://github.com/runtimeverification/mir-semantics/pull/364


sskeirik (approved):
I didn't try to build the flake --- because I'm not up to speed on the K nix situation; everything else looks good. Re: comments on slack that we could change the organization to match that other repos, I am fine with both approaches. If we do want to reorganize the repo, I think that could also be a follow-up PR.

PR #370, opened 2024-06-20
Convert stable MIR JSON to K

Baltoli:
This PR starts the process of building a test harness and related infrastructure for the new stable MIR semantics. Most of the changes here are boilerplate related to getting tests running, but the PR also sets up a basic test case for the initial panicking test case converted by @mariaKt.

The single test added asserts that converting an example JSON program to a KMIR AST produces exactly the same result as parsing a handwritten stable MIR program using the K record syntax as initially implemented in https://github.com/runtimeverification/mir-semantics/pull/364.

Once this PR is merged, my next step will be to rework https://github.com/runtimeverification/mir-semantics/pull/369 on top of the testing harness set up here so that we can begin to write tests that perform simple executions.

### DIFF
--- a/kmir/src/kmir/convert_json/__main__.py
+++ b/kmir/src/kmir/convert_json/__main__.py
@@ -2,9 +2,6 @@ from __future__ import annotations
 
 import json
 import sys
-from pathlib import Path
-
-from pyk.ktool.kprint import KPrint
 
 from .convert import from_dict
 


dkcumming:
I am not too familiar with this but I looked through everything and it seems good. I tested locally and they tests worked. I will approve but maybe I am not the best set of eyes for this one.

dkcumming (approved):

kmir/src/kmir/convert_json/convert.py
```diff
@@ -22,31 +20,57 @@ def _unimplemented() -> NoReturn:
...
+        return KApply('.maybeBytes', ())
+
+    return KApply('maybeBytes', (maybe_byte_from_dict(js[0]), bytes_from_dict(js[1:])))
 
-    return KApply('___TYPES_Bytes_MaybeByte_Bytes', (maybe_byte_from_dict(js[0]), bytes_from_dict(js[1:])))
+
+def provenance_map_entry_from_dict(js: Sequence[object]) -> KApply:
+    match js:
+        case [int(size), {'Memory': dict()}]:
+            # TODO: we actually want to parse a provenance map here; need to
```
mariaKt, line 38: As you noticed, there is a discrepancy here, between the expected AST and the serialized output. My guess is that this may be due to the `allocId` being materialized, while the AST representation expects it to be interned. I am not sure what needs to be updated here (the output of the compiler or the AST). but we will make sure to address this.

mariaKt (approved):

PR #379, opened 2024-06-27
Updated json->K according to new json format {crates,global allocations}

mariaKt:
This PR changes the json->K script to be able to parse and generate the AST for the json format as generated for the panic.rs example from the current serialization, i.e.
```
{
    "crates": {
        "name":
        "items:" {
            "name": ...
            "id": ...
            "kind": ...
            "body": ...
            "promoted": ...
            "details": (currently assumed null)
        }
        ... (other fields are currently matched but ignored)
    },
    "gallocs": []
}
```
We make small changes to the semantics (constructors with the appropriate fields for the Sorts that are included in this AST), and reference is updated to what the expected AST is.


sskeirik (approved):
Looks good! Just so you know, there are a few other changes coming soon --- sorry for the turbulence --- we're settling down soon.

PR #392, opened 2024-07-24
Unique IDs for functions and alloc IDs

mariaKt:
This PR makes changes to the manual parser script according to the output of PRs [here](https://github.com/runtimeverification/smir_pretty/pull/4) and [here](https://github.com/runtimeverification/smir_pretty/pull/5#pullrequestreview-2195054613).

The most important changes are:
- The functions map is used to replace every use of the function throughout the file with a sha256 encoding of its name. The map is then not passed through to the K AST.
- The crate_id (64 but unique ID for the crate) is used to create a unique ID for all alloc IDs in the crate. This is done by creating a 128-bit integer, appending the crate ID with the alloc ID represented as 64-bit byte strings.


kmir/src/kmir/convert_json/convert.py
```diff
@@ -98,12 +110,33 @@ def constant_kind_from_dict(js: str | Mapping[str, object]) -> KApply:
...
+            _raise_conversion_error('')
+
+
 def mirconstid_from_dict(n: int) -> KApply:
     return KApply('mirConstId(_)_TYPES_MirConstId_Int', (KToken(str(n), KSort('Int'))))
 
 
 def mirconst_from_dict(js: Mapping[str, object]) -> KApply:
     match js:
+        case {'kind': 'ZeroSized', 'ty': int(ty), 'id': int(id_)}:
```
sskeirik, line 135: So, it is possible that there are Zero Sized constants which do not refer to function types --- in that case, we want to leave those constants alone. We can check for such constants by examining whether their type id is present in our function map.

kmir/src/kmir/convert_json/convert.py
```diff
@@ -7,8 +7,20 @@
     from collections.abc import Sequence, Mapping
     from pyk.kast.inner import KInner
 
+import hashlib
+
 from pyk.kast.inner import KApply, KSort, KToken
 
+func_map: dict[int, Mapping[str, str]] = {}
```
sskeirik, line 14: Since this is a global, can we update its name to call attention to that fact? My preference is for SCREAMING_SNAKE_CASE, but other styles could work well too.

sskeirik (commented):
Looks good overall. I found one inconsistency between the parser and my understanding of the specification. After addressing that, I'd love to merge. Thanks!

kmir/src/kmir/convert_json/convert.py
mariaKt: I renamed both using this convention. I could try to thread these through from the top function to anywhere needed if you don't like the globals.

sskeirik: I don't think that's necessary --- just a visual indicator is plenty.

kmir/src/kmir/convert_json/convert.py
mariaKt: I didn't know that, thanks for point it out! Addressed.

PR #394, opened 2024-07-26
Add initial implementation of core MIR semantics

sskeirik:
This PR adds an initial implementation of the core rules of the MIR semantics. Some parts are left undone --- I have tried to add comments to call attention to these points.

NOTE: The changes made to K record field accessors were done because I was running into annoying parse ambiguities when they were present.


kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
+  rule #evalTyConst( tyConstKindZSTValue( _TY )           ) => #makeStructValue( 0, .List )
+  rule #evalTyConst( tyConstKindValue( TY, ALLOC )        ) => #loadAllocation( TY, ALLOC )
+  // NOTE: tyConstKindUnevaluated( DEF, ARGS ) should not occur --- all type system constants must be fully evaluated
+
+  syntax Value ::= #getArg( List )     [function]
+                 | #getFstArg( List )  [function]
+  syntax List ::= #getRestArgs( List ) [function]
+
+  rule #getArg( ListItem( Val:Value ) )     => Val
+  rule #getFstArg( ListItem( Val:Value ) )  => Val
```
mariaKt, line 608: `getArg` and `getFirstArg` are exactly the same, is this intentional? What is the use case of each? Is `getArg` for the single item of a list of one element and `getFirstArg` for the first item of a list of any number of elements?

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
+  rule <k>                    #eval(rvalueAddressOf(_MUT, PLACE))      => #makePtr(PLACE)            ... </k>
+  rule <k> #operands(ARGS) ~> #eval(rvalueAggregate(KIND, _))          => #makeAggregate(KIND, ARGS) ... </k>
+  rule <k> #operands(ARGS) ~> #eval(rvalueBinaryOp(OP, _, _))          => #evalOp(OP, ARGS)          ... </k>
+  rule <k> #operands(ARGS) ~> #eval(rvalueCheckedBinaryOp(OP, _, _))   => #evalOp(OP, ARGS)          ... </k>
+  rule <k> #operands(ARGS) ~> #eval(rvalueCast(KIND, _, _TY))          => #evalCast(KIND, ARGS)      ... </k>
+  rule <k> #operands(ARGS) ~> #eval(rvalueCopyForDeref(_))             => #value( #getArg(ARGS) )    ... </k>
+  rule <k> #operands(ARGS) ~> #eval(rvalueDiscriminant(_))             => #evalDisc(ARGS)            ... </k>
+  rule <k> #operands(ARGS) ~> #eval(rvalueLen(_))                      => #evalLen(ARGS)             ... </k>
+  rule <k>                    #eval(rvalueRef(_REG, _BRW_KIND, PLACE)) => #makePtr(PLACE)            ... </k>
+  rule <k> #operands(ARGS) ~> #eval(rvalueRepeat(_, _))                => #makeArray(ARGS)           ... </k>
```
mariaKt, line 624: Bug :This should create a count sized array of elem, not an array with two elements elem, count.

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
+  rule #evalArgs(nonDivergingIntrinsicCopyNonOverlapping(ARGS))                      => #evalArgs(ListItem(src(ARGS)) ListItem(dst(ARGS)) ListItem(count(ARGS)), .List)
+  rule #evalArgs(terminatorKindSwitchInt(ARG, _TARGETS))                             => #evalArgs(ListItem(ARG), .List)
+  rule #evalArgs(terminatorKindResume)                                               => .K
+  rule #evalArgs(terminatorKindAbort)                                                => .K
+  rule #evalArgs(terminatorKindReturn)                                               => .K
+  rule #evalArgs(terminatorKindUnreachable)                                          => .K
+  rule #evalArgs(terminatorKindDrop(_PLACE, _TARGET, _ACT))                          => .K
+  rule #evalArgs(terminatorKindCall(FUNC, ARGS, _DEST, _TARGET, _ACT))               => #evalArgs(#argsToList(FUNC ARGS, .List), .List)
+  rule #evalArgs(assert(COND, _EXPECTED, _MSG, _TARGET, _ACT))                       => #evalArgs(ListItem(COND), .List)
+  rule #evalArgs(terminatorKindInlineAsm(_TMPL, _ARGS, _OPTS, _SPNS, _TARGET, _ACT)) => .K
```
mariaKt, line 543: I think we need a rule like these `#evalArgs(terminatorKindGoto) -> .K`, otherwise we will get stuck when trying to evaluate the exec rule for the goto terminator.

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
+  rule <k> #exec(statementKindStorageLive(_Local))                           => .K                                               ... </k>
+  rule <k> #exec(statementKindStorageDead(_Local))                           => .K                                               ... </k>
+  rule <k> #exec(statementKindRetag(_RetagKind, _Place))                     => .K                                               ... </k>
+  rule <k> #exec(statementKindPlaceMention(_Place))                          => .K                                               ... </k>
+  rule <k> #exec(statementKindAscribeUserType(_Place, _UserType, _Variance)) => .K                                               ... </k>
+  rule <k> #exec(statementKindCoverage(_Coverage))                           => .K                                               ... </k>
+  rule <k> #exec(statementKindConstEvalCounter)                              => .K                                               ... </k>
+  rule <k> #exec(statementKindNop)                                           => .K                                               ... </k>
```
mariaKt, line 674: Could you add comments next to the exec rules for which of these statements are actually NoOps (so the .K should remain) or TODOs?

kmir/src/kmir/kdist/mir-semantics/kmir.k
mariaKt, line 543: Could you add comments next to the evalArgs rules for which of these need additional handling (TODOs) or are handled (should remain .K)?

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
-                          | Ptr( Int, Value )
-                          | Array( Bytes )
-                          | Tuple( Values )
-                          | Closure( Int, Values )
-                          | Adt( Int, Int, Values )
-                            // Semantics Specific Types
-                          | Uninit()
+  configuration <memoryConfig>
+                  <body> .BasicBlocks </body>
+                  <stack> .List </stack>
```
mariaKt, line 34: I would like to confirm with you, what is the index of the top of the stack? From what I read in the semantics, it seems to be 0. I refer to the K List data structure that we use as the stack, not the index of the base or local address that it holds. Followup: just to confirm: it seems that the indices in base, local addresses are the opposite? Is this so? Why?

kmir/src/kmir/kdist/mir-semantics/ty.k
```diff
@@ -121,6 +121,7 @@ syntax PolyFnSig ::= BinderForFnSig
 // Not needed this way. We could just do PolyFnSig ::= binderForFnSig(value: FnSig, boundVars: BoundVariableKindList).
 
 syntax Ty ::= ty(Int, TyKind) [symbol(ty)]
```
mariaKt, line 123: Is the ty(Int, TyKind) constructor no longer used?
If so, could you remove it? This would include moving the symbol to the ty(Int) constructor, changing this line (that handles creating the type K AST) in convert.py
https://github.com/runtimeverification/mir-semantics/blob/788088af85870d8807a48bd28fff341fb449239b/kmir/src/kmir/convert_json/convert.py#L446
with
```
return KApply('ty', (KToken(str(n), KSort('Int'))))
```
and removing the second argument from the `ty(...)` constructors in the reference.kmir, keeping only the `Int`.

mariaKt (changes requested):

kmir/src/kmir/kdist/mir-semantics/kmir.k
sskeirik: So, the current intended behavior is that we access the stack by counting from the bottom, not from the top.

Thus, if the stack height is N, then the bottom stack value has index 0 and the top value as index N-1.

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
 
-  syntax KItem ::= #evaluating( Operands, Values )
-                 | #evaluated( Values )
+  syntax AddressBase ::= AddrBase( Int )                 // global index
+                       | AddrBase( Int, Int )            // stack frame index, local var index
 
-  rule <k> #evaluating( Opr Oprs,  Vals )         => #evaluate( Opr ) ~> #evaluating( Oprs, Vals ) ... </k>
-  rule <k> Val:Value ~> #evaluating( Oprs, Vals ) => #evaluating( Oprs, Val Vals )                 ...  </k>
-  rule <k> #evaluating( .Operands, Vals )         => #evaluated( #concatValues(Vals, .Values) )    ... </k>
+  syntax Address ::= LocalAddr( Int,                     // stack frame height
```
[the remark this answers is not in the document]
sskeirik: In this case, `height` and `index` refer to the same thing. I can pick one for consistency.

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
+
+  rule #validSlice(Mem, From, To, Dir) => 0  <=Int From
+                                  andBool 0  <=Int To
+                                  andBool To <=Int size(Mem)
+                                  andBool 0  <=Int #sizeSlice(Mem, From, To, Dir)
+
+  rule #sizeSlice(Mem, From, To, Dir) => #if Dir
+                                           #then (size(Mem) -Int To) -Int From
+                                           #else To -Int From
+                                         #fi
```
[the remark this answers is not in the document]
sskeirik: That is the intended semantics.

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
+        ~> #prepareCall( #makeFnPtrValue( FnAddr ), .List, place(local(0), .ProjectionElems), noBasicBlockIdx, unwindActionTerminate )
+        ~> #setProcRetVal
+           ...
+       </k>
+       <globals> .Map => Mem </globals>
+       <allocs> .Map => Allocs </allocs>
+
+  // Lookup Fn Type
+
+  syntax Int ::= #lookupFnByType( Int ) [function]
```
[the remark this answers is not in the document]
sskeirik: This should have been removed due to changes in how we handle linking.

kmir/src/kmir/kdist/mir-semantics/kmir.k
sskeirik: This is a bug. Will fix.

sskeirik: So, in general, we need to eval anything that is loaded from memory. I will add comments to this effect as well.

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
            ...
        </k>
-       <callStack> ListItem(CallRecord( FnAddr( CallerAddr ), Target, Dest, _Unwind, CallerLocals )) Stack => Stack </callStack>
-       <locals> (RetValue _Locals:Values) => #setValue( CallerLocals, Dest, RetValue ) </locals>
-       <body> _CalleeBody => CallerBody </body>
-       <memory> ... CallerAddr |-> FnVal( CallerBody, _, _ ) ... </memory>
-    requires Target =/=K noBasicBlockIdx
 
-  syntax KItem ::= panicked() [symbol(panicked)]
+  rule <k> #evalArgs( .List, ARGS ) => #operands( ARGS ) ... </k>
```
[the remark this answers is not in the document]
sskeirik: Yes, this needs to be reversed. I will reverse it.

kmir/src/kmir/kdist/mir-semantics/kmir.k
sskeirik: This is a bug. `getFirstArg` should get the head of a list --- which it currently doesn't do.

kmir/src/kmir/kdist/mir-semantics/kmir.k
sskeirik: Yes, `#makeArray` does the wrong thing --- it needs to be corrected.

kmir/src/kmir/kdist/mir-semantics/kmir.k
sskeirik: Yes, I will add a comment about that.

kmir/src/kmir/kdist/mir-semantics/ty.k
sskeirik: We are not using either constructor of `ty` at the moment in the semantics, so we can skip this for now and handle updating the parser later.

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
+                 | #evalTyConst( TyConstKind )            [function]
+
+  rule #evalMirConst( _TY, _CID, constantKindZeroSized                                                     ) => #makeStructValue ( 0, .List )
+  rule #evalMirConst(  TY, _CID, constantKindAllocated( ALLOC )                                            ) => #loadAllocation( TY, ALLOC )
+  rule #evalMirConst(  TY, _CID, constantKindUnevaluated( unevaluatedConst( DEF, _ARGS, PROMOTED_INDEX ) ) ) => #loadUnevalConst( TY, DEF, PROMOTED_INDEX )
+  rule #evalMirConst( _TY, _CID, constantKindTy( tyConst( KIND, _ID ) )                                    ) => #evalTyConst( KIND )
+  // Handle extended function constant types
+  rule #evalMirConst( _TY, _CID, constantKindFnDef( Addr )                                                 ) => #makeFnPtrValue( Addr )
+  rule #evalMirConst( _TY, _CID, constantKindNoOp                                                          ) => #getNoOpFnAddr()
+  rule #evalMirConst( _TY, _CID, constantKindIntrinsic( Sym )                                              ) => #getIntrinsicAddr( Sym )
```
sskeirik, line 597: Try to add a comment about these --- about how they work.

kmir/src/kmir/kdist/mir-semantics/kmir.k
```diff
@@ -14,167 +15,717 @@ module KMIR-SYNTAX
...
+    => #if #NoOpProjection(Prj)
+         #then #set( Addr, Prjs, Scrutinee, Val )
+         #else #if Prj ==K projectionElemDeref
+           #then #setPtr( #readAddr( Scrutinee ), Prjs, Val )
+           #else #inj( Addr, Prj, Scrutinee, #set( Addr, Prjs, #prj( Prj, Scrutinee ), Val ) )
+       #fi #fi
+
+  rule #set( Addr, .ProjectionElems, _Scrutinee, Val ) => MemWrite( Addr, Val )
+
+  rule #setPtr( Addr, Prjs, Val ) => #set( #base( Addr ), #catPrjs( #getProj( Addr ), Prjs ), #loadValue( #delProj( Addr ) ), Val )
```
sskeirik, line 390: Double-check that this `#setPtr` does the correct thing.

kmir/src/kmir/kdist/mir-semantics/ty.k
mariaKt: As long as we don't use it in the execution semantics, the update is not needed for now. The current convert.ty may be deprecated soon anyway. We can handle it in a separate PR.

mariaKt (approved):

PR #397, opened 2024-08-03
Update syntax for parser

mariaKt:
This just merges some changes made to the syntactic rules in order to make them more amenable to parsing. They are mostly applied in body.k as a starting point, but they should also be applied to the other syntax. Specifically:

- Group information: Syntactic productions meant to be detected by the parser are generally in `[group(mir)]`, unless there is a more specific category: `mir-int`, `mir-string`, `mir-bool` (for productions of a single non terminal primitive sort), `mir-enum` (for productions corresponding to en enumeration), `mir-list` (for lists of sorts), `mir-option` (for sorts of options, usually named Maybe*), and specifically `mir-option-{int, string, bool}` for options of primitive sorts.
- The `---` separator follows the mir construct type if the corresponding json construct has field names. The field names follow, separated by `--`. Within a field name, `_` is represented by `-`.
- Productions should not directly include Primitive sorts unless it is a single non terminal. That is why `MIRInt`, `MIRBool`, `MIRString` were introduced. This has already been generally applied, but I make a note in case anything has been missed.
- There should not be an intermediate sort between the sort corresponding to en enumeration and the constructors.


sskeirik (approved):
Looks good to me! Thanks.

PR #401, opened 2024-08-09
Parser usability and bug fixes

mariaKt:
This PR makes the following changes related to the parser:
Ease of use:
- Adds the ability to use auto-generated klabel instead of requiring all mir productions to have symbols. Symbols **have** to be added only to rules where we need to be able to automatically construct the symbol (enums, terminals)
- Changes the interface to need the expected Sort name instead of the symbol of the top level production for the parsed term.
- Adds a testing path that uses kparse to generate the ast from the provided reference KMIR. To write a test, we need to provide the reference KMIR (instead of the KInner representation), the expected sort, and the input json.

This PR also includes additional test cases and some bug fixes that were identified using some of these tests.
- Missing `'None'` in expected Json.
- Lists were generated in reverse.
- Enums with a single non-named argument were treated as if the json object that is the value of their enumerator field name was part of that enum instead of part of its argument.

Follow ups:
- Additional changes to the semantics, in order to apply the parser's assumptions to the rules that are encountered. This can be at an as needed basis. See the notes.md that accompanies the parser for a description.
- We can have MIRInt, MIRBool, MIRString defined as
```
syntax MIRInt ::= mirInt(Int)          [group(mir-int), symbol(MIRInt::Int)]
                | Int
...
```
adjust `#getInt`, `#getStr`, `#getBool` in kmir.k as
```
rule #getInt(I:Int)       => I
...
```
and modify the parser to special handle them to generate Int, Bool, String.
- To handle K Lists from the parser, we will need to add special case handling for the rules including them. We will need to:
  - be able to identify the productions that need special handling, so that we know the intended sort of the List argument (symbol, klabel). Alternatively, we can find a way to also include those to the group somehow. For the next part, I assume that we have managed to find the sort of the List elements (e.g., Value)
  - Edit `_parse_mir_nonterminal_json` as follows: We traverse and parse the rest of the arguments normally. If/When we get to a `List` sort, instead of calling the `_parse_mir_json`, directly call a new function, `_parse_mir_klist_json`, similar to `_parse_mir_list_json`, that accepts an element_sort (instead of the list sort like `_parse_mir_list_json`) as a parameter instead of directly inferring it from the list sort. This function will be made to return a K list application, by applying the symbols for append and empty for a K list, and wrapping the parsed elements with `ListItem`.


sskeirik (approved):
Looks good to me! Thanks!

PR #402, opened 2024-08-09
Parser generates Primitive Sort instead of MIR{Primitive Sort}.

mariaKt:
This PR
- updates the MIR semantics, including the primitive Sorts `Int`, `Bool`, `String` in the definitions of `MIRInt`, `MIRBool`, `MIRString`
```
syntax MIRInt ::= mirInt(Int)          [group(mir-int), symbol(MIRInt::Int)]
                | Int
...
```
- adjusts `#getInt`, `#getStr`, `#getBool` in kmir.k as
```
rule #getInt(I:Int)       => I
...
```
in order to be able to always extract the primitive sort
- modifies the parser, adding special handling of these cases to generate `Int`, `Bool`, `String` tokens instead of `MIRInt`, `MIRBool`, `MIRString`. No injection is needed at the K AST, it is introduced as needed in KORE.

I added a test for a body. It has field `argCount`, which is a `MIRInt`, and thus exercises this path.


sskeirik (approved):
Perfect. Thanks for this!

PR #408, opened 2024-08-22
Add annotations and tweaks to parse a complete smir json

jberthold:
With these changes, the `smir_pretty` `panic_example` output json can be converted to a `Pgm` in `kast` format.

```shell
$ time poetry run -C kmir/ -- convert-from-definition ../smir_pretty/panic_example.smir.json Pgm > /dev/null && echo "Parse successful"

real	0m1.414s
user	0m1.352s
sys	0m0.062s
Parse successful
```

How can I add the test to [the `schema-parse` tests](https://github.com/runtimeverification/mir-semantics/tree/master/kmir/src/tests/integration/data/schema-parse)? 
I am not sure how to create the `kmir` file with the expected output, as that one would need to be parsed by `kparse`, rather than being the `pretty_print` output of the `kast` AST...


kmir/src/kmir/convert_from_definition/__main__.py
```diff
@@ -19,7 +19,9 @@ def main() -> None:
     tools = semantics()
     p = Parser(tools.definition)
 
-    result = p.parse_mir_json(json.loads(args.json), args.sort)
+    with open(args.json, 'r') as f:
+        json_data = json.load(f)
+    result = p.parse_mir_json(json_data, args.sort)
```
jberthold, line 24: Working with files is imperative as soon as we get to reasonable program sizes... we can always use `/dev/stdin` otherwise. 
My workflow to debug/test this was 
```
$ jj -p some.path.#.into.the..1.json.structure < input.json | \
    poetry run -C kmir -- convert-from-definition /dev/stdin SomeSort
```
We could set up tests that way, too.

kmir/src/kmir/kdist/mir-semantics/ty.k
```diff
@@ -46,283 +51,332 @@ syntax MIRString ::= mirString(String) [group(mir-string), symbol(MIRString::Str
...
+
+// syntax Promoted ::= promoted(Int) [group(mir-int)]
+syntax MaybePromoted ::= promoted(Int) [group(mir-option-int)]
+                       | "notPromoted" [group(mir-option)]
+
+syntax Align ::= align(Int) [group(mir-int), symbol(align)]
+
+// FIXME provSize and allocId are in a _list_ in json
+syntax ProvenanceMapEntry ::= // provenanceMapEntry(provSize: MIRInt, allocId: AllocId) [group(mir), symbol(provenanceMapEntry)]
+                              List [group(mir-klist-MIRInt)]
```
jberthold, line 259: I had to change this to a simple list to make it parse. Unsure whether this data structure is used much during execution...

kmir/src/kmir/kdist/mir-semantics/ty.k
```diff
@@ -46,283 +51,332 @@ syntax MIRString ::= mirString(String) [group(mir-string), symbol(MIRString::Str
...
-syntax PlaceholderForBoundRegion ::= placeholderForBoundRegion(universe: UniverseIndex, bound: BoundRegion)
-syntax RegionKind ::= regionKindReEarlyParam(EarlyParamRegion)
-                    | regionKindReBound(DebruijinIndex, BoundRegion)
-                    | "regionKindReStatic"
-                    | regionKindRePlaceholder(PlaceholderForBoundRegion)
-                    | "regionKindReErased"
-syntax Region ::= region(kind: RegionKind)
+// the struct variant of Ty was not observed in the wild yet, commenting it out
+syntax Ty ::= // ty(MIRInt, TyKind) [group(mir-enum), symbol(Ty::ty)]
+              ty(Int)            [group(mir-int)]
```
jberthold, line 148: Do we need the `ty(MIRInt, TyKind)` production at all? The `panic_example` test data did not contain any, and the parser does not support parsing either an `Int` or an object.

kmir/src/kmir/kdist/mir-semantics/alloc.k
```diff
@@ -12,15 +15,28 @@ module ALLOC
...
+                                           | "noBinderForExistentialTraitRef"                               [group(mir-option)]
+
+syntax GlobalAllocKind ::= globalAllocFunction(Instance)
+                             [group(mir-enum), symbol(GlobalAllocKind::Function)]
+                         | globalAllocVTable(Ty, MaybeBinderForExistentialTraitRef)
+                             [group(mir-enum), symbol(GlobalAllocKind::VTable)]
+                         | Static(StaticDef)
+                             [group(mir-enum), symbol(GlobalAllocKind::Static)]
+                         | Memory(Allocation)
+                             [group(mir-enum), symbol(GlobalAllocKind::Memory)]
```
dkcumming, line 30: I think the reason that this was called `GlobalAlloc` and not `GlobalAllocKind` is that it is specifically referencing [GlobalAlloc in the Stable MIR AST](https://github.com/runtimeverification/rust/blob/85f90a461262f7ca37a6e629933d455fa9c3ee48/compiler/stable_mir/src/mir/alloc.rs#L8-L22). I think the intention was to keep the non-terminals sorts the same, but I don't know if we have already deviated or if we ought to.

kmir/src/kmir/kdist/mir-semantics/ty.k
mariaKt: Maybe not yet, but it contains information that should be needed. The size is a bad name (named as in rustc internals again(, but it is really an offset, into the allocation with that allocId.

kmir/src/kmir/kdist/mir-semantics/ty.k
mariaKt: There was a large discussion about this earlier. The one we actually need is the `ty(Int)`. Initially, we included the TyKind as part of ths Ty constructor in order to materialize the interned information. Eventually this was dropped, but the constructor remained, unused.
I think that it should become `ty(MIRInt) though)`

kmir/src/kmir/kdist/mir-semantics/alloc.k
```diff
@@ -12,15 +15,28 @@ module ALLOC
...
+
+syntax GlobalAlloc ::= globalAllocEntry(MIRInt, GlobalAllocKind)
+         [symbol(globalAllocEntry), group(mir)]
+
+syntax GlobalAllocs ::= List {GlobalAlloc, ""}
+         [symbol(globalAllocsMap), terminator-symbol(.globalAllocsMap), group(mir-list)]
+
+////////////////////////////// unused for parsing?
+
```
mariaKt, line 39: This was used to hold the list stored in the allocs field of the json input. I don't know if this has been removed in more recent versions.
If it is needed, the list symbols should be updated: `GlobalAllocs::append`, `GlobalAllocs::empty`

kmir/src/kmir/kdist/mir-semantics/alloc.k
mariaKt: Yes, that was the case.

kmir/src/kmir/kdist/mir-semantics/ty.k
```diff
@@ -46,283 +51,332 @@ syntax MIRString ::= mirString(String) [group(mir-string), symbol(MIRString::Str
...
+
+syntax Tys ::= List {Ty, ""}     [group(mir-list)]
```
mariaKt, line 150: List symbols missing. I noted one, but there are others. Generally, this will not make parsing fail, unless the specific rule is needed, so it can be an ongoing process.

mariaKt (commented):
I believe that the changes made are consistent with what is needed by the parser. I did not check *every* rule for the correctness of field names or symbols, but the logic of how the mir group is selected, and field names and symbols constructed seems correct generally, except list symbols.
I think it looks good, but I am going to leave the approval to someone familiar with the current format of the json output in case there are changes I am not familiar with.

kmir/src/kmir/kdist/mir-semantics/ty.k
jberthold: OK so it looks like I can just delete it

kmir/src/kmir/kdist/mir-semantics/alloc.k
jberthold: Well... but we have the convention for `mir-list` that the list sort is the plural of the non-list sort.
Maybe we can just use untyped K-lists in the semantics, though, then we can rename things freely.

kmir/src/kmir/kdist/mir-semantics/alloc.k
jberthold: Indeed, we have to align those symbols with the rest, thanks.

jberthold: As before, `GlobalAllocs` holds the data from the `allocs` field in the json, it was just renamed. The `unused` comment was referring to the syntax below, I have removed it (if this is dead code we clean it up later).

kmir/src/kmir/kdist/mir-semantics/alloc.k
mariaKt: That's true!

In any case, all these names were given as they are in the Stabe MIR AST, and then any additional structures that were created were made to not clash.
Then, when the parser was written, we created a collection of guidelines that needed to be followed so that we could identify which rule we need to apply. If there is the need to change names to achieve that, this would take priority.

I am sure that it is possible to change the names of the sorts that we created that were not directly in the Stable MIR AST in order to keep the ones that are the same, if this is something that you think is important. The reasoning behind doing it this way initially was to easily know which sort corresponds to which part of the AST.

mariaKt:
I just saw the question about creating the reference. I was just writing it manually, working from the bottom up (e.g., create tests for some `StatementKind` before `Statement`, some `Statement` before `Statements`, ...).

dkcumming (approved):

dkcumming:
Jost and I talked through this in a meeting! In order to move forward we are pushing ahead! @sskeirik your feedback is still appreciated when you have time

PR #409, opened 2024-08-22
Documentation that links to the source code

dkcumming:
This PR converts the K source files to markdown files so that we can annotate the files with an index that links to Stable MIR, SMIR, Internal MIR. While this will inevitably get stale at some point, it may be handy to help people get familiar with the complicated connections between the different modules and hopefully speed things up overall.


jberthold (approved):
I think this is a good start to set the scene for adding documentation incrementally into the K files, so I'd say let's merge it _now_.
You did well _moving_ most of the files (except the `abi.{k,md}`, which is fine), that way ~the changes won't conflict~. EDIT: there will only be conflicts with `requires "someFile.k"` directives at the top of the K files.

PR #413, opened 2024-08-28
411 simple integration test using old simple rust programs

jberthold:
This PR adds a set of simple rust programs (from prior art) for a simple parser test:
 * salvages [old test programs from `run-rs` in the `textual-mir` branch](https://github.com/runtimeverification/mir-semantics/tree/textual-mir/kmir/src/tests/integration/test-data/run-rs)
 * [sets up `smir_pretty`](https://github.com/runtimeverification/smir_pretty) as a submodule for integration tests
 * adds `make` targets to manage the submodule build (including a bootstrap of `rustc`, which [should be set up as a recursive submodule](https://github.com/runtimeverification/smir_pretty/issues/8) in the future)
 * adds a `make` target for a simple integration test: convert each `run-rs` program to Stable-MIR and parse the result.

Remarks: 
* The tests are currently only proving that the parser succeeds. We can add better checks later when the format is stable enough to not cause noise in golden tests.
* In order to run `smir_pretty`, information about the rust target architecture is needed. To circumvent this, the related script (within `smir_pretty` repo) is patched (but only the copy in the docker file system).  This is a workaround and could be addressed properly by
  - crafting a github action to set up K and removing the dockerised K,
  - or preparing `smir_pretty` to run within `nix` and using a nix shell.
  - or (not recommended) placing the custom `rust` build and `smir_pretty` within the docker image or container.


.github/workflows/test.yml
```diff
@@ -76,3 +76,62 @@ jobs:
...
+
+      - name: 'HACK: patch rustc_arch.sh script (within docker)'
+        run: |
+          arch=$(rustc -vV | sed -n -e 's/host: \(.*\)$/\1/p')
+          docker exec --user user mir-smir-ci-${GITHUB_SHA} \
+              bash -c "printf '#!/bin/sh\necho \"$arch\"\n' > deps/smir_pretty/rustc_arch.sh"
+          docker exec --user user mir-smir-ci-${GITHUB_SHA} \
+              cat deps/smir_pretty/rustc_arch.sh
+          docker exec --user user mir-smir-ci-${GITHUB_SHA} \
+              deps/smir_pretty/rustc_arch.sh
```
dkcumming, line 130: @jberthold This is the hack to get our version of rustc running for smir pretty on CI?

jberthold: Yes. The problem is that we need to run inside the docker image (to have `kompile`) but there is no `rustup` in the image. That script uses `rustup` to find `rustc`, then calls `rustc` just to find out the target architecture string (which is part of the path to the libraries to link to). :curly_loop: :loop: 
The hack is to find that architecture string _outside the docker image_ and then patch the script _inside_ the docker image's file system (which is not shared with the host for some other reason... :shrug: )  to print that string

dkcumming: Okay my understanding is that this will not need to rebuild the stage 2 compiler from our semantics from consecutive runs of the tests, and even if the K source is changed as the semantics will kompile but the stage 2 compiler is already built.

dkcumming (approved):

dkcumming:
I will let you choose when to merge @jberthold in case there was more commits to add

.github/workflows/test.yml
jberthold: Yes, I even think the entire set of compiler artefacts could be cached, and tried to set up a cache for it, but the compiler build setup is multi-stage and therefore more complex to avoid rebuilds. For now this was just supposed to be the first step - our setup with the compiler dependency is definitely "suboptimal" anyway (and should/will change soon).

PR #436, opened 2025-01-27
Make `kmir run` parse in smir pretty jsons

gtrepta:


jberthold (approved):
LGTM.
We need to do something about error handling at some point, but let's focus on actually _running_ now.

dkcumming (approved):
:+1:  I agree with Jost

ehildenb:
Nice! Also would be cool to start some little testing harnesses that enable running little MIR blobs and seeing the output, and have an associated "failing" list that we can update as we get more stuff passing.

PR #440, opened 2025-01-30
Add --depth option to `kmir run`

gtrepta:
Just a quick little addition that allows us to use `--depth` with `kmir run`, helps with debugging.


jberthold (approved):
LGTM

PR #441, opened 2025-01-31
Load functions array into config

gtrepta:
TODO: Lookup in the functions array when handling a call terminator

Stretch goal: Fuse the `<functions>` and `<function-names>` cells together.


kmir/src/kmir/kdist/mir-semantics/kmir.md
```diff
@@ -91,6 +91,12 @@ module KMIR-CONFIGURATION
                   // FIXME where do we put the "GlobalAllocs"? in the heap, presumably?
                   // function store, Int -> Body
                   <functions> .Map </functions>
+                  <function-names>
+                    <function-name multiplicity="*" type="Map">
+                      <ty> 0 </ty>
```
gtrepta, line 96: This might need to be `MIRInt` and not just `Int`. I didn't run into an issue here when I was testing this though.

PR #443, opened 2025-01-31
Updating smir_pretty to stable-mir-json

dkcumming:
`https://github.com/runtimeverification/smir_pretty/` has been renamed to `https://github.com/runtimeverification/stable-mir-json` so the submodule needs to be updated to the right url


.github/workflows/test.yml
```diff
@@ -79,7 +79,7 @@ jobs:
 
   smir-integration-tests:
     needs: code-quality-checks
-    name: "Integration with smir_pretty"
+    name: "Integration with stable-mir-json"
```
jberthold, line 82: This confuses github... the old name was a required check for PRs. Can disable it , then merge and then enable the requirement for the new name.

jberthold (approved):
LGTM. Need to change the `Required checks for PRs`  option to merge it.

dkcumming:
> LGTM. Need to change the `Required checks for PRs` option to merge it.

I think I configured it back to the way it ought to be

PR #445, opened 2025-02-04
Basic support for control flow (calls and returns without data)

jberthold:
* Various adaptations to `kmir.md` so that we can look up functions by their type `Ty`
  - `functions` table is now keyed on `Ty` instead of `DefId`
  - `caller` and `currentFunc` fields likewise contain a `Ty`
  - the `main` function is added to the functions table with `ty(-1)` key to work around a shortfall of stable-mir-json: https://github.com/runtimeverification/stable-mir-json/issues/42
  - various other adaptations and stub code to write function call arguments to resp. local values
* A basic program where `main()` calls `a()`, which calls `b()`, which calls `c()` is added as a first integration test
* The `#init` function is adapted to support executing functions other than `main` (not tested yet)


gtrepta (approved):
LGTM. I opened #446 to merge into this branch. Feel free to merge in my changes. Otherwise, I'll go over your review when I get back to work tomorrow.

PR #446, opened 2025-02-04
`--start-symbol` for `kmir run`

gtrepta:
This adds a `<start-symbol>` cell to the configuration with a configuration variable for the symbol to begin execution from.

It also changes the expected output in the `exec-smir` integration tests to only check the contents of the K cell. The semantics are going to be very unstable right now, and without this change we will probably be spending time updating the test vectors on nearly every pull request.


kmir/src/kmir/__main__.py
```diff
@@ -60,6 +56,9 @@ def _arg_parser() -> ArgumentParser:
     run_parser = command_parser.add_parser('run', help='run stable MIR programs')
     run_parser.add_argument('input_file', metavar='FILE', help='MIR program to run')
     run_parser.add_argument('--depth', type=int, metavar='DEPTH', help='Depth to execute')
+    run_parser.add_argument(
+        '--start-symbol', type=str, metavar='SYMBOL', default='main', help='Symbol name to begin execution from'
+    )
```
jberthold, line 61: :+1: good idea.
I was struggling to find a way to have a _default value_ for a K config variable (there is probably none), this is much easier.

kmir/src/kmir/kdist/mir-semantics/kmir.md
```diff
@@ -92,6 +92,7 @@ module KMIR-CONFIGURATION
                   // heap
                   <memory> .Map </memory> // FIXME unclear how to model
                   // FIXME where do we put the "GlobalAllocs"? in the heap, presumably?
+                  <start-symbol> $STARTSYM:String </start-symbol>
```
jberthold, line 95: Can we maybe just make this cell have sort `Symbol` and insert `symbol("$STARTSYM")`? (instead of adding `symbol(_)` in the code below)

kmir/src/tests/integration/test_integration.py
```diff
@@ -187,14 +187,23 @@ def test_exec_smir(
...
+                kcell = term
+
+    collect(get_k_cell, kast_result)
 
     with output_kast.open('r') as f:
         expected = f.read().rstrip()
 
-    result_pretty = tools.kprint.kore_to_pretty(result).rstrip()
+    assert isinstance(kcell, KApply)
+    result_pretty = tools.kprint.pretty_print(kcell).rstrip()
```
jberthold, line 207: Not sure we should give up on comparing all of the configuration.
How about we add a `make` target to both re-generate the `*.smir.json` and  the expected output?
The expected output in my version is easy to produce just by piping `kmir run` output into a file, and we could run `stable-mir-json` before so we also update the json. This would be a standard task to run locally as a preparation when opening  a PR.

gtrepta: A make target to update expected outputs will in certain circumstances be a good tool to have.

But, what's the point of having these tests, if they change with every change we make? I think we should try for something that's at least semi-stable so we don't have as many larger irrelevant diffs in our PRs.

jberthold: Well... if the test fails, or it you proactively change the expectation file using that `make` target, reviewers are encouraged to look at the changed output. We can still all choose to skip those changes in the test expectations but they provide examples of how the configuration is expected to change with the respective code change in the  PR.
If you think that's useless we can of course also just scrape the tests, or compare selected cells. But I thought it is acceptable to maintain and useful. Let's see what @dkcumming thinks

dkcumming: I think it is nice to check the diffs, at least that way we can tell if things are changing as we expect. But this is assuming there isn't much overhead. I think it is preferable to have these extra steps if they are manageable though - preferable to not noticing we changed something spuriously and finding it later and having to retrace steps.

gtrepta: Ok, I've added `make update-exec-smir` and reverted the test outputs. I do believe that the K configuration exposes implementation details that we should prefer not to have and should focus on interfacing with it to test only behavior (ie. function return values, successful execution, anything we can model about Rust's semantics), but we can kick that can down the road. I wont push for it here.

deps/stable-mir-json
jberthold, line 1: This was in one of my commits on the other PR by mistake.
Maybe best to `rebase  -i master` your current branch and remove all commits from the other PR. Or else you can of course revert the submodule downgrade manually.

jberthold (approved):
Apart from the submodule downgrade, this LGTM, thanks for adapting.

deps/stable-mir-json
gtrepta: Thanks for catching this. When I merged `master` in I had too many conflicts that didn't make sense so I just used the "ours" strategy, this submodule downgrade must have slipped in that way.

PR #451, opened 2025-02-10
Merge `kmir/Makefile` into `./Makefile`, adapting the CI workflow

jberthold:
All targets from `kmir/Makefile` were ported (i.e., just copied, actually) into the top-level `Makefile`.

The new default target is `build`, which installs the python code and builds the K definition.

The `test-integration` target (previously in `kmir/Makefile`) will now check that the K definition is up to date before testing, to avoid a stale `definition.kore`. `test-unit` does not, in order for CI to not require a K framework to run it.

Running the global `make test` will execute both the python (unit and integration) tests and the `smir-parse-tests`. The latter still do _not_ depend on `build` because we want to run them in CI with a docker-supplied `K` toolkit and `pyk` without running the entire test within docker (because it also needs a Rust toolchain).


Makefile
```diff
@@ -18,8 +17,8 @@ stable-mir-json:
 
 # generate smir and parse given test files (from parameter or run-rs subdirectory)
 smir-parse-tests: TESTS = $(shell find $(PWD)/kmir/src/tests/integration/data/run-rs -type f -name "*.rs")
-smir-parse-tests: SMIR = cargo -Z unstable-options -C deps/stable-mir-json run -- 
-smir-parse-tests:
+smir-parse-tests: SMIR = cargo -q -Z unstable-options -C deps/stable-mir-json run --
```
jberthold, line 20: Added a `-q` here so we don't get cluttered output per test.
(when debugging problems we can ovveride the variable)

dkcumming (approved):

PR #458, opened 2025-02-18
Modularize Value representations

gtrepta:


kmir/src/kmir/kdist/mir-semantics/kmir.md
```diff
@@ -217,7 +217,7 @@ be known to populate the `currentFunc` field.
   rule <k> #execFunction(
               monoItem(
                 SYMNAME,
-                monoItemFn(_, _, body(FIRST:BasicBlock _ #as BLOCKS,LOCALS, _, _, _, _) .Bodies)
+                monoItemFn(_, _, body((FIRST:BasicBlock _) #as BLOCKS,LOCALS, _, _, _, _) .Bodies)
```
jberthold, line 220: Weird... I remember having to modify this `#as` expression before to address the problem of `currentBody` ending up empty _on my machine_ but not on CI! Maybe this slipped back to prior code somehow, or it was a different place.
Would be great to understand what causes the difference - the K version should be the same.

jberthold (approved):
LGTM, let's get this in and I will continue the "HIGH" one on top of this change.

PR #459, opened 2025-02-19
Match on sorted entries instead of use lookups + projections

gtrepta:
This changes a few rules that use lookups on collections to instead match on them in a more structural manner.

These changes are all in `kmir.md`, there's one more in there that's a little more involved which could use more investigation. There are also lots more in `rt/data.md`, but I have another PR open that would create a lot of conflicts with changes made there, so I'm waiting to do that.


dkcumming (approved):
LGTM

jberthold (approved):
LGTM, definitely nicer than before.
~I might try to modify the operations in `rt/data.md` in a similar way some time soon~ EDIT: Just re-read the description, you probably have those changes somewhere locally...

Issue #457, opened 2025-02-18
`kmir prove` on k claims and `krun` on symbolic input


gtrepta:
A `kmir prove` that can run some basic k claims of the mir-semantics can help us understand how execution on the symbolic backends performs and behaves.
A prerequisite for this, a `kmir run` which operates on symbolic input using HS backend and RPC interface will be helpful on its own.

PR #466, opened 2025-02-26
457 - Add `kmir gen-spec` command

gtrepta:
Progress for #457 

This adds a command `gen-spec` which takes an SMIR json file and generates a specification module with a claim that rewrites the initial configuration to the `#EndProgram` symbol on the k cell.


jberthold (approved):
LGTM

PR #468, opened 2025-02-28
Add `--haskell-backend` to kmir run

gtrepta:
This adds the option `--haskell-backend` to `kmir run` to make it easier to test on the command line.

This also runs the haskell backend on the `exec_smir` tests in addition to the llvm backend.


kmir/src/kmir/__main__.py
```diff
@@ -45,7 +46,8 @@ def __init__(self, input_file: Path, output_file: str | None, start_symbol: str)
 
 
 def _kmir_run(opts: RunOpts) -> None:
-    tools = semantics()
+    print(f'haskell_backend: {opts.haskell_backend}')
```
jberthold, line 49: This will probably go away before merging?
```suggestion
```

jberthold (approved):
LGTM (with the `print` removed)

dkcumming (approved):

kmir/src/kmir/__main__.py
gtrepta: Yep, that's some debugging left over.

PR #472, opened 2025-02-28
457 - kmir prove

gtrepta:
fixes: #457 

This adds the `kmir prove` command which can load `.k` files containing specification modules and run the prover on the claims contained within them. It has support for `--proof-dir` to save the KCFGs as well as `--include-labels` and `--exclude-labels` to specify which claims to run if you don't want to run all of them.


jberthold (approved):
LGTM

kmir/src/kmir/__main__.py
```diff
@@ -69,23 +90,39 @@ def _kmir_gen_spec(opts: GenSpecOpts) -> None:
...
-    new_k_cell = tools.kmir.Symbols.END_PROGRAM
+    new_k_cell = KMIR.Symbols.END_PROGRAM
     rhs = CTerm(Subst({'K_CELL': new_k_cell})(config_with_cell_vars))
-    claim, _ = cterm_build_claim(opts.input_file.stem, lhs, rhs)
+    claim, _ = cterm_build_claim(opts.input_file.stem.replace('_', '-'), lhs, rhs)
 
     output_file = opts.output_file
     if output_file is None:
-        output_file = opts.input_file.with_suffix('.spec.json')
+        suffixes = ''.join(opts.input_file.suffixes)
```
jberthold, line 103: Small Q: Maybe I misunderstand the python code here, but should this maybe rather be `'.'.join`?

kmir/src/kmir/__main__.py
```diff
@@ -45,6 +47,25 @@ def __init__(self, input_file: Path, output_file: str | None, start_symbol: str)
...
+    exclude_labels: tuple[str, ...] | None
+
+    def __init__(
+        self, spec_file: Path, proof_dir: Path | str | None, include_labels: str | None, exclude_labels: str | None
+    ) -> None:
+        self.spec_file = spec_file
+        if proof_dir is None:
+            self.proof_dir = None
+        else:
+            self.proof_dir = Path(proof_dir).resolve()
```
dkcumming, line 64: I think this can be the same format as the lines below no?
```suggestion
        self.proof_dir = Path(proof_dir).resolve() if proof_dir is not None else None
```

dkcumming (approved):
I think this is all correct. Maybe some good proofs to test it out would be some internal functions like the recent `BinOp` stuff Jost added. But I think this is good to merge as is

kmir/src/kmir/__main__.py
gtrepta: The elements in `suffixes` contain the dots.

PR #477, opened 2025-03-05
kmir `prove view` for viewing proofs.

gtrepta:
More for #457 

Adds subcommands for `kmir prove`. `kmir prove run` was already implemented in #472. This implements `kmir prove view` which opens the TUI explorer on a proof.


kmir/src/kmir/__main__.py
```diff
@@ -70,6 +71,13 @@ def __init__(
         self.bug_report = bug_report
 
 
+@dataclass
+class ProveViewOpts(KMirOpts):
+    spec_file: Path
+    proof_dir: Path | None
+    id: str
```
ehildenb, line 78: Given that `spec_file` and `proof_dir` are both common among `ProveRunOpts` and `ProveViewOpts`, you may considre factoring out a `KMirProveOpts` which has the common arguments (and handling/defaults), and then specialize them more in the class.

You can do the same with `argparse` commands, by using the `parents` field to inherit options between different parsers.

PR #478, opened 2025-03-06
Refactor `TypedLocal` to lift `Moved` and `NoValue` higher

jberthold:
The previous `TypedLocal` used `MaybeValue` with the following definition

```
syntax MaybeValue ::= Value | "NoValue" | "Moved"
```

As it turned out when writing proof claims, the use of `Value` without a wrapping constructor here creates issues; an injections would have to be added but is not. Also, `TypedLocal` instances carrying a `Moved` value were carrying redundant type and mutability information.

Therefore, the distinction between current, moved, and uninitialised values was lifted to `TypedLocal` which now has three variants

```
syntax TypedLocal ::= typedLocal ( Value, MaybeTy, Mutability ) | Moved | noValue( Ty, Mutability )
```

The `MaybeTy` construction might turn out to be similarly problematic in the future but is left in place for now.


dkcumming (approved):
LGTM

PR #479, opened 2025-03-06
Adding multi-crate support through cell map

dkcumming:
Work In Progress

Support for multi-crate Rust programs. This PR uses a cell map to store the crates. Another option would be to join all of the `Items` into a giant mono-crate. There is potential that using cell maps may be challenging for symbolic execution, and we should consider the other approach.

This PR:
- Changes `Pgm` to `Crate`
- Parsing and Interpreting now can take multiple `.smir.json` files as input
- Files are parsed into a K List
- When interpretting, that list is processed to populate a cell map of each crate
- TODO: mkFunctions
- TODO: Multi crate call
- TODO: Examples / tests


jberthold:
This approach is now obsolete since we are linking the SMIR JSON ahead of execution

PR #484, opened 2025-03-12
Load proof from proof dir when viewing

jberthold:
The previous code was creating a new proof object from the spec file (first claim). Instead, the proof needs to be loaded from a file in the proof directory.

Note that `APRProver.read_proof` and `APRProver/read_proof_data` have different expectations to how the proof is stored. The latter method has to be used.

* `kmir prove view` now does not require a spec file argument any more.
* In turn, the `--proof-dir` argument is now _required_ (unsaved proofs cannot be viewed)


dkcumming (approved):
LGTM

Issue #480, opened 2025-03-07
Remove old parser


dkcumming:
Two attempts at parsing were created, however the `v2parser` in `convert-from-definition` is the only one that has been continuously updated. The old version should be removed and references to v2 removed as it will be the only parser from then on.

jberthold:
Completed in #489

PR #489, opened 2025-03-13
Refactor, deps update, and README additions

### LINKED ISSUE #481: Update `README.md`
The `README.md` is lagging behind the features that we have implemented, it should be updated to reflect the current state of KMIR.

dkcumming:
- Removed the old parser, and `convert-from-json`. The V2 parser in `convert-from-definition` is the only parser now.
- Added parsing to `README.md`
- Updated `stable-mir-json`

Closes #481 #480


dkcumming (commented):
I haven't done this, but I think it would be better if we remove the `convert-from-defintiion` directory and move the logic for parsing inside `KMir` so we can have a `kmir parse`. Would you agree that is an improvement @guy @jberthold ?

gtrepta:
> I haven't done this, but I think it would be better if we remove the `convert-from-defintiion` directory and move the logic for parsing inside `KMir` so we can have a `kmir parse`. Would you agree that is an improvement @guy @jberthold ?

I think the logic for parsing in smir definitely should stay in its own module/file. But, rearranging the CLI tools so `convert-from-definition` is just `kmir parse` sounds good to me.

jberthold:
> > I haven't done this, but I think it would be better if we remove the `convert-from-defintiion` directory and move the logic for parsing inside `KMir` so we can have a `kmir parse`. Would you agree that is an improvement @guy @jberthold ?
> 
> I think the logic for parsing in smir definitely should stay in its own module/file. But, rearranging the CLI tools so `convert-from-definition` is just `kmir parse` sounds good to me.

Agree, let's keep it separate but use a common CLI.

And BTW that "guy" is not our guy @gtrepta  :smile:

kmir/src/kmir/convert_from_definition/parser.py
jberthold, line 1: Maybe better to `git rm ...parser.py` and then `git mv ...v2parser.py ...parser.py` so we keep history.
Not too important this time, I guess.

jberthold (approved):

kmir/src/kmir/convert_from_definition/parser.py
dkcumming: Yes that is what I did. This is what you mean right?
image

dkcumming:
Okay I will change it the CLI and keep it separate before I merge

PR #498, opened 2025-03-19
486 support writing to dereferenced values

### LINKED ISSUE #486: Support for references and dereferencing
Many stdlib functions use (shared or mutable) references to values.

These references will refer to locals in dfifferent stack frames
* the reference value needs a stack depth (initially 0) for access
* stack depth needs to be updated when passing references as arguments or result
* we need to clarify whether and under what conditions references can be written to.

jberthold:
Writing to (mutable) references is now supported, required a refactoring of how projections are handled for writes.

Closes #486 

Also formats the JSON of reference-related tests from the previous PR and reduces the amount of test proofs to run.


dkcumming (approved):
A great improvement

PR #501, opened 2025-03-21
Refactor evaluation rules to use more specific sorts

jberthold:
* introduces heating/cooling for `#setLocalValue`
* introduces sort `Evaluation` with subsorts `TypedValue` (result type), `Projected`, and `RValue`
* We might also add `Operand` later and remove `#readOperand` in favour of more heating/cooling rules.
* uses `Evaluation` as result sort of `#compute` and `#arithmeticInt` (hopefully avoids excessive branching on symbolic values)


jberthold:
OUtdated

PR #505, opened 2025-03-24
Single `Body` `Instance`s

dkcumming:
[Stable MIR JSON PR#60](https://github.com/runtimeverification/stable-mir-json/pull/60) removes the error where we thought we had to find the `Body` for promoted items. This is reflecting that change by removing the list of `Body` and changing it to an `maybeBody` for `MonoItem::Fn` since sometimes a `Body` is not present


kmir/src/tests/integration/data/exec-smir/arithmetic/arithmetic-unchecked-runs.smir.json
jberthold, line 1: Can we please pretty-print the checked-in JSON (to get a chance to read a small diff)?

kmir/src/tests/integration/data/exec-smir/arithmetic/arithmetic.smir.json
jberthold, line 1: Ditto, please reformat using jq. Something like 
`for file in $(find ...exec-smir -name \*smir.json); do echo $file;  jq . $file > $file.pretty && mv $file.pretty $file; done`

jberthold (approved):

kmir/src/tests/integration/data/exec-smir/arithmetic/arithmetic-unchecked-runs.smir.json
dkcumming: 100% I should of done that

kmir/src/tests/integration/data/exec-smir/arithmetic/arithmetic.smir.json
dkcumming: I don't think the tab size was consistent with all of these. They are all formatted with `jq .` now so hopefully things are consistent from now on

jberthold: Quite possibly I used vscode's formatter on some of them before...

PR #511, opened 2025-03-27
506 refactor value handling

jberthold:
* distinguish `TypedLocal` by subsorts `TypedValue`, `NewLocal`, `MovedLocal` (renaming constructors)
* introduce supersort `Evaluation` with result sort `TypedValue` for `Operand` and `RValue`
* introduce `MIRError` sort inhabited by all errors, and a labelled stop rule
* use heating/cooling for a number of constructs that read or/and write locals


kmir/src/kmir/kdist/mir-semantics/kmir.md
```diff
@@ -435,14 +429,33 @@ NB that a stack height of `0` cannot occur here, because the compiler prevents l
...
+         <dest> _ => NEWDEST </dest>
+         <target> someBasicBlockIdx(TARGET) => NEWTARGET </target>
+         <unwind> _ => UNWIND </unwind>
+         <locals> ListItem(_:NewLocal) _ => NEWLOCALS </locals>
+       //</currentFrame>
+       // remaining call stack (without top frame)
+       <stack> ListItem(StackFrame(NEWCALLER, NEWDEST, NEWTARGET, UNWIND, NEWLOCALS)) STACK => STACK </stack>
+       <functions> FUNCS </functions>
+     requires CALLER in_keys(FUNCS)
+     [preserves-definedness] // CALLER lookup defined
```
dkcumming, line 458: I am trying to think of what this looks like in the mir code. What is an example of this?

dkcumming (approved):
Looks great

kmir/src/kmir/kdist/mir-semantics/kmir.md
jberthold: The new rule is for the case where local `_0` has not been written to (it is still sort `NewLocal` ) but the function returns. That happens regularly when a function returns nothing (i.e., unit). We used to "write" the `noValue` as a NoOp but now the sort is different so we should skip it.

PR #515, opened 2025-03-28
Refactor split RT-DATA into separate modules/files

jberthold:
* A basic module `rt/value.md` contains the `Value` sort and `TypedLocal` with its subsorts
* A `rt/numbers.md` module contains code related to decoding and casting integers, and helper functions
* The configuration now lives inside `rt/configuration.md`

Most of this is just moving existing code around into new files and adjusting imports.

The `RT-DATA-LOW` and `RT-DATA-HIGH` modules were removed, moving code from `HIGH` into `RT-DATA` (`RT-DATA-LOW` was never actually used).

## New module import graph (omitting `domains.md` modules):

image

We can split `rt/data.md` into more parts, e.g.
* [ ] reading/writing locals and projections, 
* [ ] `Rvalue` evaluation (in general), which could also move to `KMIR-CONTROL-FLOW` inside `kmir.md`
* [ ] `BinaryOp`, `UnaryOp` (which are specific `Rvalue`s)


dkcumming (approved):
Lgtm

PR #520, opened 2025-04-04
Extend type metadata from stable mir json

jberthold:
Upgrades the `stable-mir-json` dependency and adjusts parser and execution to use the new `Basetype` and `TypeInfo` sorts.

Second change to implement `enum` support will follow this one in a separate PR.


kmir/src/kmir/kdist/mir-semantics/ty.md
```diff
@@ -219,6 +219,13 @@ syntax ExistentialPredicateBinders ::= List {ExistentialPredicateBinder, ""}
   syntax Movability ::= "movabilityStatic"  [group(mir-enum), symbol(Movability::Static)]
                       | "movabilityMovable" [group(mir-enum), symbol(Movability::Movable)]
 
+  // additional sort for all base types in RigidTy
```
dkcumming, line 222:
```suggestion
\```
`BaseType`s are not present as an explicit type in MIR or Stable MIR AST, but are a conceptual sub-sort of `Ty` and `RigidTy` for types that are "primitive".
```k
```
I believe we should we should be clearer about what a base type is since it is our construction but relates to an idea in MIR and Stable MIR.

(The backslash is there because I had trouble with the formatting of the suggestion)

dkcumming: I even wonder if we should change the name from `BaseType` to `PrimitiveType`, what do you think?

dkcumming (approved):

kmir/src/kmir/kdist/mir-semantics/ty.md
jberthold: Sure, I can change the name. Will do that, and add your comment there, too

PR #523, opened 2025-04-07
Added a feature checklist.

dkcumming:
This list is conservative and many things I have labelled as "Not supported" if I am not sure how this will be in the code.


docs/feature-checklist.md
```diff
@@ -0,0 +1,422 @@
+Instead of rewriting our own feature checklist from scratch, we make a derivative of the
+[Verus checklist](https://github.com/verus-lang/verus/blob/main/source/docs/guide/src/features.md)
+which is available for use under [MIT](https://github.com/verus-lang/verus/blob/main/LICENSE) 
+license.
```
jberthold, line 4: How about we make that a html comment? 
```suggestion
<!-- Instead of rewriting our own feature checklist from scratch, we make a derivative of the
[Verus checklist](https://github.com/verus-lang/verus/blob/main/source/docs/guide/src/features.md)
which is available for use under [MIT](https://github.com/verus-lang/verus/blob/main/LICENSE) 
license.
-->
```

docs/feature-checklist.md
```diff
@@ -0,0 +1,422 @@
...
+
+**Last Updated: 2025-04-07**
+
+<div class="table-wrapper"><table>
+  <thead><tr><th colspan="2"><strong>Items</strong></th></tr></thead>
+  <tbody>
```
jberthold, line 10: I would much prefer for this table to be a set of markdown tables. We don't need the HTML code and it will get in the way when we update the table.
What cannot be done is the merged cells as sub-headers. we can do separate tables for each as one option

dkcumming:
@jberthold I think all your items from your feedback are addressed! Want to sign off and perhaps add automerge if you are happy?

docs/feature-checklist.md
```diff
@@ -0,0 +1,140 @@
...
+|Feature|Status|
+|-------|------|
+|Functions, methods, associated functions|Supported|
+|Associated constants|Not supported|
+|Structs|Partially supported|
+|Enums|Partially supported|
+|Const functions|Not supported|
+|Async functions|Not supported|
+|Macros|Not supported|
+|Type aliases|Not supported
```
jberthold, line 19: I think we do support macros and type aliases - by way of using MIR which is way later than macro expansion and type checking. Not sure for type aliases (whether they can occur as a `TyKind` variant in the phase we feeed on) but definitely we have no problem nor business with macros.

Maybe a category `N/A (MIR)` would be in order. We could also merge and amend this table later, though.

docs/feature-checklist.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Struct/enum definitions
+|Feature|Status|
+|-------|------|
+|Type parameters|Not supported|
+|Where clauses|Not supported|
+|Lifetime parameters|Not supported|
+|Const generics|Not supported|
+|Custom discriminants|Not supported|
+|Public / private fields|Not supported|
```
jberthold, line 31: Again because the compiler has resolved it for us before MIR, we are probably "supporting" some of these.
Custom discriminants are definitely  supported as of today.
```suggestion
## Struct/enum definitions
|Feature|Status|
|-------|------|
|Type parameters|Not supported|
|Where clauses|Not supported|
|Lifetime parameters|Not supported|
|Const generics|Not supported|
|Custom discriminants| Supported|
|Public / private fields|N/A (MIR)|
```

docs/feature-checklist.md
```diff
@@ -0,0 +1,140 @@
...
+|`await`|Not supported|
+|Unsafe blocks|Supported|
+|`&`|Supported|
+|`&mut`, place expressions|Supported|
+|`==`, `!=`|Supported|
+|Type cast (`as`)|Partially supported|
+|Compound assigments (`+=`, etc.)|Supported|
+|Array expressions|Not supported|
+|Range expressions|Not supported|
+|Index expressions|Supported|
```
jberthold, line 54:
```suggestion
|Index expressions|Not supported (Arrays not supported)|
```

docs/feature-checklist.md
```diff
@@ -0,0 +1,140 @@
...
+|Tuple expressions|Supported|
+|Struct/enum constructors|Not supported|
+|Field access|Partially supported|
```
jberthold, line 57:
```suggestion
|Field access|Supported|
```

docs/feature-checklist.md
```diff
@@ -0,0 +1,140 @@
...
+|Labels, break, continue|Supported|
+|Return statements|Supported|
+
+## Integer arithmetic
+|Feature|Status|
+|-------|------|
+|Arithmetic for unsigned|Supported|
+|Arithmetic for signed (`+`, `-`, `*`, `/`, `%`)|Supported|
+|Bitwise operations (`&`, `\|`, `!`, `>>`, `<<`)|Supported|
+|Arch-dependent types (`usize`, `isize`)|Supported|
```
jberthold, line 69:
```suggestion
|Arch-dependent types (`usize`, `isize`)|Not supported (fixed width) |
```

docs/feature-checklist.md
```diff
@@ -0,0 +1,140 @@
...
+## Types and standard library functionality
+|Feature|Status|
+|-------|------|
+|Integer types|Supported|
+|`bool`|Supported|
+|Strings|Not supported|
+|`Vec`|Not supported|
+|`Option` / `Result`|Supported|
+|Floating point|Not supported|
+|Slices|Supported|
```
jberthold, line 80:
```suggestion
|Slices|Not supported|
```

docs/feature-checklist.md
dkcumming: Yeah I also figured this to be true for macros but I also haven't didn't know if there was something I wasn't thinking of like attributes that might be unsupported. But yeah without investigating type alias I will just leave it as `Not supported` until someone has a look to convince themselves its all good

PR #524, opened 2025-04-10
Update `kmir run` to behave more like `cargo run`

gtrepta:
- Introduce `CargoProject` class which can be used to gather information about a cargo project.
- `kmir run` now uses `CargoProject` which finds the `.smir.json` artifact in the target directory after invoking `cargo build`.

Note that currently either `RUSTC` will need to be set to the `stable-mir-json` install, or the smir json files should already have been built that way and exist in the target directory. Later, we can get rid of this requirement when we package `stable-mir-json` with `kmir`.


dkcumming (approved):
LGTM!

README.md
```diff
@@ -36,7 +36,7 @@ Use `--help` with each command for more details.
 
 `parse` to parse a Stable MIR JSON file (`*.smir.json`) file to a K AST
 
-`kmir run` to load an SMIR JSON generated by the `stable-mir-json` tool.
+`kmir run` to run a binary target in your cargo project. The SMIR JSON will need to have been already built with `stable-mir-json`, or you will need the `RUSTC` environment variable set to your installation of `stable-mir-json`.
```
jberthold, line 39: Is it _required_ for `kmir run` to have a `cargo` project now?
(I don't think it is, and I would not want it that way TBH). Maybe reformulate this and give an example?

gtrepta: Yes, this makes it required. I can write a dev script that still works with a `.smir.json` argument if you're still trying to work that way. We've been discussing making the CLI interface more like using cargo, this is meant to be a step in that direction.

jberthold: I think we could have a second option (say `--file`) which, when provided, would not use the `cargo` global at all.
Not everyone who writes a small Rust program wants to, nor has to, use cargo so  I think it would make sense to keep around a way to simply run a given `*.smir.json` file.

kmir/src/kmir/__main__.py
```diff
@@ -80,7 +81,12 @@ class ProveViewOpts(KMirOpts):
 def _kmir_run(opts: RunOpts) -> None:
     tools = haskell_semantics() if opts.haskell_backend else llvm_semantics()
 
-    parse_result = parse_json(tools.definition, opts.input_file, 'Pgm')
+    cargo = CargoPackage(Path.cwd())
+
+    target = opts.bin if opts.bin else cargo.default_target
+    smir_file = cargo.smir_for(target)
```
jberthold, line 87: This is where the `--file` option would be implemented: if given, `smir_file is taken from there instead of asking `cargo`.

README.md
jberthold:
```suggestion
`kmir run` to run a binary target in your cargo project (or a provided `*.smir.json` file).
For `cargo, the SMIR JSON will need to have been already built with `stable-mir-json`,
or you will need the `RUSTC` environment variable set to your installation of `stable-mir-json`.
```

jberthold (approved):
LGTM

Issue #495, opened 2025-03-18
Add symbolic arguments  to execution according to start symbol


jberthold:
Currently only `start-symbol`s without arguments can be executed successfully.
The initial stack setup must include inspection of the function that is executed to determine what arguments should have symbolic values (and represent them as best as possible as `Value` instances).

jberthold:
Notes from discussion:
* `#reserveFor` in `kmir.md` would need to generate symbolic initialised values for the arg.s
* The function to do this might be similar to `#decodeConstant` from `rt/data.md`
* For Integer and Bool we can generate typed `Value`s, aggregates and enum values would remain entirely symbolic (missing info about component types)

gtrepta:
Pushed up https://github.com/runtimeverification/mir-semantics/tree/guy/symbolic-args branch with a [`KMIR-SYMBOLIC-LOCALS`](https://github.com/runtimeverification/mir-semantics/blob/guy/symbolic-args/kmir/src/kmir/kdist/mir-semantics/symbolic/kmir-symbolic-locals.md) module that has symbolic variants of `#init` and `#execFunction` that will set up the arguments to a function as symbolic variables.

gtrepta:
Fixed by #525

PR #525, opened 2025-04-10
Generate symbolic arguments based on function signature

gtrepta:
fixes: #495 

- Creates a symbolic `KMIR-SYMBOLIC-LOCALS` module with symbolic counterparts to the `#init` and `#execFunction` symbols in `KMIR`.
- Modifies `kmir gen-spec` to use `#initSymbolic` in the generated claim.


kmir/src/kmir/kdist/mir-semantics/symbolic/kmir-symbolic-locals.md
```diff
@@ -0,0 +1,98 @@
+# Symbolic locals for functions
+
+This module contains rules that follow the same structure as `#init` in the KMIR module, however it reads the arguments for the signature of the start symbol and generates symbolic values for them.
+
+```k
+requires "../kmir.md"
+
+module KMIR-SYMBOLIC-LOCALS [symbolic]
```
jberthold, line 8: When reading this, I liked the idea of moving all initialisation functions from `KMIR-CONTROL-FLOW` into a new module that control flow can call into.
However, we might have to keep every symbolic rule (that seems to mean involving existentials here) in its own symbolic module? 
I don't see why `#initSymbolic` needs to be different from `#init` otherwise, we simply should not execute a function that has arguments without supplying any.

kmir/src/kmir/kdist/mir-semantics/symbolic/kmir-symbolic-locals.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Declare symbolic arguments based on their types
+
+```k
+  syntax KItem ::= #reserveSymbolicsFor( LocalDecls, Int )
+
+  rule <k> #reserveSymbolicsFor( .LocalDecls, _ ) => .K ... </k>
+
+  rule <k> #reserveSymbolicsFor( LOCALS:LocalDecls, 0 ) => .K ... </k>
+       <locals> ... .List => #reserveFor(LOCALS) </locals> // No more arguments, treat the rest of the Decls normally
```
jberthold, line 53: nit-pick: Maybe call this `#reserveSymbolicArg` ?

jberthold (approved):
LGTM
It would be great to have a single `#init` function but maybe the two backends just require this split into two of them. Having a way to provide _concrete_ arguments and separate rules for the two backends in one module might enable this (later, not now).

kmir/src/kmir/kdist/mir-semantics/symbolic/kmir-symbolic-locals.md
```diff
@@ -0,0 +1,98 @@
...
+```k
+  rule <k> #reserveSymbolicsFor( localDecl(TY, _, MUT) LOCALS:LocalDecls, COUNT        )
+        => #reserveSymbolicsFor(                       LOCALS:LocalDecls, COUNT -Int 1 )
+           ...
+       </k>
+       <locals> ... .List => ListItem(typedValue( BoolVal( ?_BOOL:Bool ), TY, MUT )) </locals>
+       <types> ... TY |-> typeInfoPrimitiveType ( primTypeBool ) ... </types>
+    requires 0 <Int COUNT
+
+endmodule
```
jberthold, line 97: We also need a rule for arguments which aren't `Int` or `Bool` valued, they would just be symbolic altogether.

PR #538, opened 2025-04-29
Structured symbolic arguments

jberthold:
Implements generating symbolic arguments with all available (known) structure for custom start symbolis

TODO
* [ ] ncludes some draft code to be refactored  (should use heating/cooling)
* [ ] A simple test should be added
* [ ] struct` s are still abstracted completely because field type information is missing (`stable-ir-json` change required)


jberthold:
This code is now mostly obsolete because we are initialising symbolic variables in the python  client.
Still leaving the branch intact in case we want to go back to it.

PR #542, opened 2025-04-29
Added testing harness for proofs from rust input

dkcumming:
This PR introduces a new testing harness that for running rust proofs. In particular we:
- Introduced a new python utility `cargo_get_smir_json` that runs Stable MIR JSON through cargo on a given rust file (all very hardcoded right now)
- Created a test directory `tests/integration/data/prove-rs` of rust programs to prove do not end in error state or break assertions
- Created a testing harness `test_prove_rs` to generate a claims for the test rust files and then prove those claims

Two tests are currently added, one that expects a pass, one that expects a fail and both work. I also tested putting incorrect values in there and they did fail as expected.

**NEXT STEPS**
- Refactor the test into it's own method which the test and `gen-spec` use
- Debug the failing `sub` and `div` proofs


kmir/src/kmir/rust/cargo.py
```diff
@@ -80,3 +92,13 @@ def default_target(self) -> str:
             )
 
         return bin_targets[0]
+
+
+def cargo_get_smir_json(rs_file: Path) -> dict[str, Any]:
+    smir_json_result = SMIR_JSON_DIR / rs_file.with_suffix('.smir.json').name
+    run_process_2(['cargo', 'run', '--', '-Zno-codegen', str(rs_file)], cwd=SMIR_JSON_DIR)
+    json_smir = json.loads(smir_json_result.read_text())
```
jberthold, line 100: Remark: This does not actually have to use `cargo`, it is a single crate because it starts from a single Rust file.
We don't have linking for multi-crate programs atm, but let's still use cargo here (recompilation might become a bit tricky, though)

jberthold (approved):
As discussed, let's merge this and follow up making it a command.

PR #544, opened 2025-04-30
Docker tweaks

jberthold:
A few improvements to make the docker containers start faster (`fixuid` goes through loads of files and takes >3min to start - those were probably the CI problems with permissions that we observed when using `docker exec`)

* [X] delete a large number of files that aren't needed to work with  the docker image
* [x] relocate `rustup` installation out of `/home/kmir`
* [x] only run `fixuid` on the `kmir` home directory 

* [X] change `RUN` command to use json style (avoids warning messages from the plugin in CI)

Before this, it took very long to start the container , now the start time is down to 9 seconds on my machine.
We could try to not `chmod` in `$HOME/.cache` (where a lot of files from the K definition are sitting) but that might break the run.


ehildenb:
Overall this looks fine to me, but I don't know what this docker image is used for. So I'll approve, but if it's something you're not sure of you should maybe wait for Freeman's approval.

ehildenb (approved):

PR #545, opened 2025-04-30
Speed up CI Docker startup times

ehildenb:
Currently we're `chown`ing the entire home directory after doing `docker cp`, but only need to do so for the directory we're copying in (`workspace`), so this makes that change. This avoids the slow `chown` of the installed Rust file.

This reduced the last `docker exec` command to <5s, from where it was seemingly a minute or more before.


dkcumming (approved):
Yay!

PR #546, opened 2025-04-30
Introduced `prove-rs` command

dkcumming:
Added a command `kmir prove-rs <FILE.rs>`. Details:
- This command takes a rust source file and generates a spec that will pass a proof if it ends successfully without violating an assertion on all paths
- Constructs an `APRProver` and attempts to prove the generated spec
- Optional arg `--max-depth <INT>` to specify the number of steps taken between proof nodes
- Optional arg `--max-iterations <INT>` to specify the total number of steps the prover can take
- Optional arg `--bug-report <FILE>` to create a bug report at the chosen location
- I also changed the logic of `test-prove-rs` to call `kmir-prove-rs`, and moved it to `test_integration.py`

Created `KMIR` method `apr_proof_from_kast` to factor out creating the `APRProof` for a refactor of `gen_spec` and `prove_rs`


kmir/src/kmir/kmir.py
```diff
@@ -49,6 +54,37 @@ def kcfg_explore(self, label: str | None = None) -> Iterator[KCFGExplore]:
...
+        tools = Tools(self.definition_dir)
+        smir_json = cargo_get_smir_json(rs_file)
+
+        parser = Parser(self.definition)
+        parse_result = parser.parse_mir_json(smir_json, 'Pgm')
+        assert parse_result is not None
+        kmir_kast, _ = parse_result
+        assert isinstance(kmir_kast, KInner)
+
+        config = tools.make_init_config(kmir_kast, 'main')
```
dkcumming, line 72: Hardcoded `main` instead of start symbol

kmir/src/kmir/kmir.py
```diff
@@ -49,6 +54,37 @@ def kcfg_explore(self, label: str | None = None) -> Iterator[KCFGExplore]:
...
+
+        lhs = CTerm(config)
+
+        rhs_subst = Subst({'K_CELL': KMIR.Symbols.END_PROGRAM})
+        rhs = CTerm(rhs_subst(config_with_cell_vars))
+        kcfg = KCFG()
+        init_node = kcfg.create_node(lhs)
+        target_node = kcfg.create_node(rhs)
+        apr_proof = APRProof('PROOF', kcfg, [], init_node.id, target_node.id, {})
+        with self.kcfg_explore('PROOF-TEST') as kcfg_explore:
```
dkcumming, line 83: Hardcoded generic names

kmir/src/kmir/kmir.py
ehildenb: I think that's OK, we'll address this when we get to symbolic arguments on functions.

jberthold (approved):
LGTM. Next will be to add an option for executing not `main` but other functoins (without arg.s)

PR #547, opened 2025-04-30
Update UV to 0.7.2 in github workflow and update uv.lock file

jberthold:


.github/workflows/test.yml
```diff
@@ -40,6 +40,8 @@ jobs:
         uses: actions/checkout@v4
       - name: 'Install uv'
         uses: astral-sh/setup-uv@v5
+        with:
+          version: 0.7.2
```
ehildenb, line 44: Also modify `.github/workflows/Dockerfile` to install this version of uv too.

ehildenb:
```
curl -LsSf https://astral.sh/uv/0.7.2/install.sh | sh
```

dkcumming (approved):
lgtm

PR #548, opened 2025-05-01
Refactor options classes and CLI args

ehildenb:
This PR refactors the options classes and CLI args to deduplicate some things, and use the options classes directly when invoking the prover. In particular:

- All the classes `*Opts` are moved out of `__main__` and into `options`.
- The function `Prove.prove_rs(...)` now takes a `ProveRSOpts` argument instead of all the individual arguments on their own.
- We factor out a common `ProveOpts` class that `ProveRunOpts` and `ProveRSOpts` draw from instead of duplicating their entries.
- We factor out a common `prove_args` CLI parser that `prove_rs_args` and `prove_run_args` pull from.
- This also makes a small change where we make sure to update the `kmir/uv.lock` file on version bumps to keep that from happening locally afterward.


dkcumming (approved):
Looks great

PR #549, opened 2025-05-01
Inline `Tools` class into `KMIR` and replumb

ehildenb:
In preparation of building LHSs of specs directly, I've inlined the `Tools` class directly into the `KMIR` class, and propagated all the appropriate changes through.

This also required getting `KParse` to match the diamond import pattern of `KRun` and `KProve` (which I think is eventually the intention, to upstream it), so I did that as well.

Overall, this moves `make_init_config` and `run_parsed` directly into `KMIR`, so it's available for all commands that `KMIR` exposes.

As one minor build system improvement, I allow the user to set variable `PARALLEL` on the CLI when invoking `make` to spede things up.


kmir/src/kmir/__main__.py
```diff
@@ -29,7 +29,7 @@
 
 
 def _kmir_run(opts: RunOpts) -> None:
-    tools = haskell_semantics() if opts.haskell_backend else llvm_semantics()
+    kmir = KMIR(LLVM_LIB_DIR)
```
ehildenb, line 32:
```suggestion
    kmir = KMIR(HASKELL_DEF_DIR) if opts.haskell_backend else KMIR(LLVM_DEF_DIR)
```

kmir/src/tests/integration/test_integration.py
```diff
@@ -350,33 +348,32 @@ def test_schema_kapply_parse(
 ]
 
 
-@pytest.mark.parametrize('tools', [llvm_semantics(), haskell_semantics()], ids=['llvm', 'haskell'])
 @pytest.mark.parametrize(
     'test_case',
     EXEC_DATA,
     ids=[name for (name, _, _, _) in EXEC_DATA],
 )
```
ehildenb, line 355: This parametrization needs to be brought back somehow.

ehildenb:
Need to check that both versions of the tests (llvm and haskell) are being run for `test_exec_smir`.

jberthold (approved):
Nice clean-up!

PR #552, opened 2025-05-01
Code to write to array indexes

jberthold:


ehildenb:
Can we add the same test to `prove-rs` test harness? You can run just those tests with `make test-integration TEST_ARGS='-k test_prove_rs`

ehildenb (approved):

ehildenb:
We checked separately that the program was indeed present in the SMIR file and had not been optimized to the trivial program by the compiler.

PR #555, opened 2025-05-01
Add `kmir show ...` command, promote prove commands to top-level

ehildenb:
This PR:

- Adds `kmir show ...` (similar to `kmir view ...`), which in addition to printing the KCFG structure actually displays the entire node present at each step. We can fine-tune the display of this later.
- Promotes the `prove` subcommands to the top-level, since they work the same with both `kmir prove run ...` and `kmir prove-rs ...`. In particular we have `kmir prove run ... => kmir prove ...`, and `kmir prove prune ... => kmir prune ...`, and `kmir prove view ... => kmir view ...`.
- Another minor refactoring where the `cargo.py` module is moved out of the `rust/` submodule, as it was the only file there.


dkcumming (approved):
I tested each of these commands locally and they all worked as expected

PR #556, opened 2025-05-01
SMIRInfo helper class

gtrepta:
This should be used to get information from SMIR to create K terms on the python side of things.


dkcumming (dismissed):
Opps I approved and commented on the wrong PR!

ehildenb (approved):

PR #561, opened 2025-05-04
Switch to using version file for stable-mir-json

ehildenb:
Blocked on: https://github.com/runtimeverification/devops/pull/234

This works a bit better with our update-deps logic, and will allow us to switch to packaged versions later rather than submodule.


dkcumming (approved):
Went through this live with Everett

PR #567, opened 2025-05-06
Fix prove-rs for use in docker container and other small things

jberthold:
* Name `prove-rs` proofs after their start symbol
* In `cargo.py` to get the smir.json, use `stable-mir-json` if we are not in a source/build tree


kmir/src/kmir/cargo.py
```diff
@@ -95,8 +95,17 @@ def default_target(self) -> str:
...
+        # prefer local dependency if it exists (i.e., we are in a source/build tree)
+        command = ['cargo', 'run', '--', '-Zno-codegen', str(rs_file)]
+        cwd = IN_TREE_SMIR_JSON_DIR
+    else:
+        # otherwise use 'stable-mir-json' from the path (fail if it does not exist)
+        command = ['stable-mir-json', '-Zno-codegen', str(rs_file)]
+        cwd = Path.cwd()
+
+    smir_json_result = cwd / rs_file.with_suffix('.smir.json').name
+    run_process_2(command, cwd=cwd)
```
ehildenb, line 108: I think we shouldn't have the "prefer in-tree" solution at all actually. We should let the user decide outside the application how they want to setup their path and stuff, where they can decide which version they prefer.

ehildenb: Not a blocker for merging, but I think there's no reason to keep the old logic around.

jberthold: > I think we shouldn't have the "prefer in-tree" solution at all actually. We should let the user decide outside the application how they want to setup their path and stuff, where they can decide which version they prefer.

OK, we can remove that code then. But maybe later, when we have updated documentation, too. I just want the docker installation to work out of the box right now so IMHO the `stable-mir-json` call is the fall-back at the moment. Will follow up removing the in-tree one later.

jberthold:
> Looks good. Do any of our installs add `stable-mir-json` to the path currently?

Not that I would know, just the `Dockerfile.kmir`. Even the name `stable-mir-json` is set only in the docker installation, not by `cargo` in any way.

PR #568, opened 2025-05-07
Test suite for showing proof results

ehildenb:
This adds the ability to test the `show` output of given proofs, in order to make sure we're preserving (or advancing) proof structure on runs. In particular:

- Add testing helper `assert_or_update_show_output` which assist in either demonstrating the same output or updating existing output for stored golden files.
- Adds a pytest fixture for using the option `--update-expected-output` to testing to instruct it to update the golden files.
- Adapts the `test_prove_rs` harness to allow specifying some proofs taht we want to check the output KCFG for, and threads through the options to use `--update-expected-output`.
- Adds a test we've been using as an example which currently gets stuck (proof fails) to the `prove-rs` test-suite. This example uses a local raw pointer, which we haven't implemented.

In order to update the golden files for the tests, run:

```
make test-integration TEST_ARGS='--update-expected-output'
```


dkcumming (approved):
LGTM

PR #570, opened 2025-05-07
Refactor projection calculations

ehildenb:
This PR attempts to make projection calculations a bit simpler to follow by factoring them out into a helper function. A very systematic approach is taken, and may be easier to review commit-by-commit:

- First, the logic that was in K rules is turned into functions, by threading through the `<stack>` and `<locals>` contents as explicit arguments for a new function `makeProjectedUpdates`.
- Next, uneeded state is removed from this function, including the new value to write (which is held in the outer KItem constructor).
- Finally, we make calculating projections actually not check mutabilitiy at all. We still retain the mutability check when writing to the actual slot, but for calculating the projection/context I figure we don't need to write to any of the slots, so mutability doesn't matter. If we can come up with a test that would fail because of this, then we can revisit this change. But as far as I can tell, `#buildUpdate` checks the mutability of the locations we're writing to anyway.


ehildenb:
Subsumed for now: https://github.com/runtimeverification/mir-semantics/pull/582

PR #571, opened 2025-05-08
Unchecked arithmetic proofs and lemmas

jberthold:
The proofs for unchecked arithmetic (from AWS initiative challenge 11) require additional lemmas to confirm that no truncation is happening. These lemmas are added here, also removing one lemma that we had twice (in the lemmas file and in `numbers.md` where `truncate` is defined).

Also checking in the Rust code over which we run the proofs, and will set up a convention for which functions to run in `prove-rs` (suggestion: one-line comment containing a list of `start-symbol` parameters  at the top of the file).


kmir/src/tests/integration/test_integration.py
```diff
@@ -451,16 +451,36 @@ def test_prove_rs(rs_file: Path, kmir: KMIR, update_expected_output: bool) -> No
...
+        else:
+            start_symbols = ['main']
+
+        for start_symbol in start_symbols:
+            prove_rs_opts.start_symbol = start_symbol
+            apr_proof = kmir.prove_rs(prove_rs_opts)
+            if not should_fail:
+                assert apr_proof.passed
+            else:
+                assert apr_proof.failed
```
jberthold, line 486: One drawback of doing this is that the proofs will run sequentially one after the other, not parallel. We could duplicate the file instead of running several proofs  from the same file, though.

jberthold:
There seems to be a problem with `unchecked_mul_*` proofs, I am avoiding those for now.

kmir/src/tests/integration/data/prove-rs/unchecked_arithmetic.rs
```diff
@@ -0,0 +1,120 @@
...
+unchecked_add_claim_for!(unchecked_add_i8   , i8   );
+unchecked_add_claim_for!(unchecked_add_i16  , i16  );
+// unchecked_add_claim_for!(unchecked_add_i32  , i32  ); // already above
+unchecked_add_claim_for!(unchecked_add_i64  , i64  );
+unchecked_add_claim_for!(unchecked_add_isize, isize);
+unchecked_add_claim_for!(unchecked_add_u8   , u8   );
+unchecked_add_claim_for!(unchecked_add_u16  , u16  );
+unchecked_add_claim_for!(unchecked_add_u32  , u32  );
+unchecked_add_claim_for!(unchecked_add_u64  , u64  );
+unchecked_add_claim_for!(unchecked_add_usize, usize);
```
ehildenb, line 60: Am I reading this correctly that it's basically making a copy of the above macro for every line here?

I wonder if we could do this using a trait instead? Then make the proof over the trait, and teh prover then goes and discharges the proof obligation for each implementation of the trait (and we define the implementation for the different types we care about).

Just a thought about exploring UX possibilities, no nede to change this PR.

kmir/src/tests/integration/test_integration.py
dkcumming: I think this is fine for a first version, we can figure out how to do it more performant later

dkcumming (approved):
Need to think of some way to solve specification restrictions we currently have. But this looks great!

kmir/src/tests/integration/data/prove-rs/unchecked_arithmetic.rs
jberthold: Agree it would be great to have a polymorphic test function but there are some technical obstacles if we did that.
1) We would need to call the test function from `main` at the type we want to test, because all code not reachable from `main` gets removed from the MIR.
2) Then again, the monomorphic function to test won't have an easy name, it will be subject to name-mangling in the compiler. That naming is probably relatively stable but not intuitive to the users (therefore not a UX improvement).

PR #572, opened 2025-05-08
Do not check mutability when computing projections

ehildenb:
The current logic for computing projections looks at the mutability of fields when _reading_ them, which shouldn't be necessary as we should be allowed to read mutable fields (just not write them). So this removes the mutability check, which means the additional arguments `FORCE` and `MUT` are not needed for `#projectUpdate`. So we refactor the bits of state that aren't needed to be in a different `KItem` in the second position on the `<k>` cell.

In particular:

- Remove the checks on each rule of `#projectUpdate` which inspect the mutability of the read fields and the `FORCE` field.
- Pulls out the state that is no longer needed to actually compute the projection (the new value to write, and the `FORCE`) into a second `KItem` on the `<k>` cell, and re-wires the rules to use this when doing `#setLocalValue`.


ehildenb:
Subsumed for now: https://github.com/runtimeverification/mir-semantics/pull/582

Issue #573, opened 2025-05-08
Refactoring `#projectedUpdate`


ehildenb:
I've tried a few experiments in refactoring `#projectedUpdate` https://github.com/runtimeverification/mir-semantics/pull/572 and https://github.com/runtimeverification/mir-semantics/pull/570. Both pass the current tests, but I'm not convinced they would pass more complicated tests, in fact I'm not convinced the current algorithm would pass more complicated tests.

Specifically, if we have a `mutable` field inside an `immutable` ref, for example if using "Interior Mutability" via `RefCell`, we should allow writes to the internal field. The current logic for computing projections will stop as soon as it hits its first `immutable` marker.

Another issue I see is that `#projectedUpdate` is really doing two things, (i) it's first finding the location that should be written to, which can change via the `projectionElemDeref`. This `projectionElemDeref` is also used for `#readProjection` and basically can switch which different location (`<locals>` or `<stack>` for example) is being dereferenced. In the case we hit one of these, we actually throw out the entire context we've been computing thus far in `#projectedUpdate`, and start over on the new location.

In summary, I think that:

- We don't need to compute contexts until we're on the last segment of the `ProjectionElems`, where a "segment" is defined as section between `projectionElemDeref` markers.
- We don't need to check mutability until we're on the very last lookup, it's OK to step through immutable elements to reach the inner mutable that we are actually update. @dkcumming has offered to come up with some tests of inner mutability so we can test that (on the current implementation, and on updates).

As a refactoring, I would suggest:

- Implement `syntax MaybeValueAndContext ::= #lookupFieldWithCtx(TypedValue, ProjectionElems)`, which handles all of the `ProjectionElem`, but if it has any `ProjectionElem` after a `projectionElemDeref`, it returns a `None`. If there is a type-mismatch in the projection and the typed value it's looking at, it will also return `None`. On successful lookup, it returns both a `Value` (the contents at the lookup path), and a `Context` (the information needed to reconstruct the original surrounding value).
[...]

PR #574, opened 2025-05-08
Add tests for mutability

dkcumming:
Adding tests to examine the boundaries of mutability and immutability, not all of these use `unsafe` but only the ones I listed as unsafe do we annotate unsafe directly:
- mut_const.rs: A UB test that mutates the data under an immutable value through pointer arithmetic
- interior-mut-fail.rs: Safely has interior mutability through `RefCell` (stuck on `Rvalue::AddressOf`)
- interior-mut2-fail.rs: Safely has interior mutability through `Cell` (stuck on `Rvalue::AddressOf`)
- interior-mut3-fail.rs: Unsafe but sound interior mutability through `UnsafeCell` (stuck on `Rvalue::AddressOf`)

We discussed the possibility of adding tests that show demonstrate immutable field of mutable structs, but this is not possible in rust. There is a notion of private fields if structs come from other modules, but this is not the same as mutability.


kmir/src/tests/integration/test_integration.py
```diff
@@ -436,9 +436,7 @@ def test_prove(spec: Path, tmp_path: Path, kmir: KMIR) -> None:
 
 PROVING_DIR = (Path(__file__).parent / 'data' / 'prove-rs').resolve(strict=True)
 PROVING_FILES = list(PROVING_DIR.glob('*.rs'))
-PROVE_RS_SHOW_SPECS = [
-    'local-raw-fail',
-]
+PROVE_RS_SHOW_SPECS = ['local-raw-fail', 'interior-mut-fail', 'interior-mut2-fail', 'interior-mut3-fail']
```
ehildenb, line 439: We should do these one per line instead. Will make it easier to add/delete them.

ehildenb:
Can we rebase this on master to get current `master` behavior? I want to open a PR with the refactoring from https://github.com/runtimeverification/mir-semantics/issues/573), and make sure we all agree on the changes that happen then.

kmir/src/tests/integration/test_integration.py
dkcumming: Can do

gtrepta (approved):

PR #575, opened 2025-05-08
Adding machine info

dkcumming:
The `MachineInfo` semantics productions are now connected to the JSON which now includes it with an update. This is ignored in the `Pgm` currently, but could be used to make `isize` and `usize` dynamic on `pointer-width` - although, I am not sure how to do that atm


ehildenb:
We shouldn't merge with it pointing to a different branch of stable-mir-json, that should be merged first (which I think is not a breaking change here?)

Also, what are we using the machine info for? Specifically for constants? maybe we should process those in the Python first, so that we only build K terms that have the higher-level data present in them?

dkcumming:
I only had it pointing at a branch while I was figuring out how to get it working.
It is needed for knowing the bit width of a `usize` and `isize` as well. But I brought it into the K config for two reasons:
1. There were already the productions for it in `target.md`
2. I thought I would be able to add it to the K configuration and make `usize` and `isize` work with the appropriate bitwidth there instead of hard coding it to `64`.

I couldn't see an easy way to achieve 2. (@gtrepta also helped and couldn't see a neat solution). But with all the parsing connected I figured I would open the PR. Do you think there is a way of getting around the `usize` and `isize` with python also?

gtrepta (approved):

PR #578, opened 2025-05-09
Added `nullaryOpAlignOf` for primitives

dkcumming:
`nullaryOpAlignOf` for primitives:
- Returns correct values for `u{8,16,32,64,128}`, `i{8,16,32,64,128}`, `bool`, `char`, `floats`;
- Assumes `usize` and `isize` are `64bit`;
- Hardcoded return size of `64bit` since `isize` and `usize` assume `64bit` currently;

We already have tests that trigger this however they are going to require `thunk` to be implemented to get past the casts.

Also I feel alignment should be platform specific, so I added a FIXME because I believe this values are essentially assumptions. Or am I misunderstanding?


gtrepta (approved):
LGTM. I think the FIXMEs are appropriate for now.

PR #581, opened 2025-05-12
Use helper for updating expected output

ehildenb:
This changes the test-suite for `test_exec_smir` to use the helper `assert_or_update_show_output` for updating expected output the same as how is done for the `prove_rs` test-suite. This means you can't use the environment variable anymore for this, instead:

```
make test-integration TEST_ARGS='--update-expected-output'
```


dkcumming:
I tested this and it worked as expected. Perhaps there is also some merit to considering including `stable-mir-json` to generate the json also.

PR #584, opened 2025-05-13
Initialise a call config in python (with symbolic arguments)

jberthold:
`make_call_config` produces a configuration with a `Call` terminator for the given start symbol in the K cell.
Arguments to the function will be created as locals of a dummy caller's stack frame, and copied by `OperandCopy` in the `Call` terminator.

Arguments are symbolic values, with as much structure as the type metadata for the argument types allows.

* Primitive types are created with variables for the values. For `Integer`, constraints limit the size of the value.
* `struct`s and `enum`s are created as `Aggregate` values but with a single variable (list-sorted) for the arguments. With more metadata for structs, we could create the struct fields with their resp. types.
* Tuples are created with a list of their components (recursively structured)
* References are created by creating the pointee value _behind_ the argument list

The dummy caller's locals have the following structure:

```
  Return    | Function arguments           | Pointees for
  value     | (symbolic structured values) | ref. arguments
------------------------------------------------------------
NewValue(0) | ARG_1 | ARG_2 | ... | ARG_N  | P_1 | P_2 | ...
-------------------------------------------|----------------
                       |                           A
                       \__(offset 0, place(N+2)____/
```
For sake of an example, ARG_2 here is a reference to data that was (symbolically) created at location P_2, behind the locals that serve as arguments of the called function. When the called function's locals are set up, the reference in `ARG_2` is copied and its offset adjusted to 1.


kmir/src/kmir/kmir.py
```diff
@@ -73,28 +75,86 @@ def make_init_config(
...
+                    types,
+                ),
+            ),
+            'ADTTOTY_CELL': KApply(
+                'mkAdtMap',
+                (
+                    map_empty(),
+                    types,
+                ),
+            ),
```
ehildenb, line 116: Each of these can probably be a single line, no need to amke it take up so much vertical space.

I would make them a single line each, and then use `make format` to decide which needs to be broken. The formatter should make the decisions about line lengths.

kmir/src/kmir/kast.py
```diff
@@ -0,0 +1,210 @@
...
+        self.pointees = []
+        self.constraints = []
+        self.counter = 1
+        self.ref_offset = 0
+
+    def run(self, local_types: list[dict]) -> tuple[list[KInner], list[KInner]]:
+        self.ref_offset = len(local_types) + 1
+        for ty in [t['ty'] for t in local_types]:
+            self._add_local(ty)
+        return (self.locals + self.pointees, self.constraints)
```
ehildenb, line 131: If you call this method `def __call__(...)`, you can do `ArgGenerator(smir_info)(local_types)` directly. See https://github.com/runtimeverification/k/blob/db2706d64499fda526e2a7b374bf24c32436e7c3/pyk/src/pyk/cterm/cterm.py#L354 for an example. Stylistic choice.

ehildenb (dismissed):
Looks really good to me overall! Some things it makes me think about:

- Probably `parser` submodule should be directly importing and using the `SMIRInfo` class, rather than us passing in the JSON directly to it. The parser can then use the extracting helpers to get the info it needs for parsing.
- I'm not super happy with how stateful `ArgGenerator` is, but it seems like it's totally encapsulated by `symbolic_locals` as a function, so it should be OK as long as we don't start passing around `ArgGenerator`s as arguments to things.

kmir/src/kmir/kast.py
```diff
@@ -21,7 +22,7 @@ def int_var(var: KVariable, num_bytes: int, signed: bool) -> tuple[KInner, Itera
     bit_width = num_bytes * 8
     var_max = ((1 << (bit_width - 1)) if signed else (1 << bit_width)) - 1
     var_min = -(1 << (bit_width - 1)) if signed else 0
-    constraints = (leInt(var, token(var_max)), leInt(token(var_min), var))
+    constraints = (mlEqualsTrue(leInt(var, token(var_max))), mlEqualsTrue(leInt(token(var_min), var)))
     term = KApply('Value::Integer', (var, token(bit_width), token(signed)))
     return term, constraints
```
ehildenb, line 27: Hmmmm, why didn't CI catch this error? Do we have any tests exercising these cases?

ehildenb:
@jberthold can you add tests of the new functionality? Specifically, of calling a function with arguments? You can add it to the `fail` and `show` lists.

kmir/src/kmir/kast.py
gtrepta: I don't think any of our prove_rs integration tests are exercising functions with parameters. I'll add some that do.

gtrepta: Correction: we DO have tests that work with parameterized functions.

Before fixing the `start_symbol` issue in my commit, `main` was just being called in all of the prove-rs tests, so the `mlEqualsTrue` issue was hidden.

All should be good.

kmir/src/kmir/kmir.py
jberthold: I agree - the formatter did this vertical layout, actually...

kmir/src/kmir/smir.py
```diff
@@ -61,9 +62,180 @@ def function_arguments(self) -> dict[str, dict]:
...
+        res = {'main': -1}
+        for item in self._smir['items']:
+            if not SMIRInfo._is_func(item):
+                continue
+
+            mono_item_fn = item['mono_item_kind']['MonoItemFn']
+            name = mono_item_fn['name']
+            sym = item['symbol_name']
+            if not sym in fun_syms:
+                continue
```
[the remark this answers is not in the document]
jberthold: It is common that some items are not in the `fun_syms`, `main` is the prime example (not sure about others, TBH)
`fun_syms` (and the `functions` table in the JSON) only contains functions that are _called_ from within the program.

kmir/src/tests/integration/data/prove-rs/show/symbolic-args-fail.expected
jberthold, line 1: This output does not really show the generated arguments (except the one `ARG_BOOL3` for the branch) but there is no good way to print the locals without getting non-deterministic IDs from `stable-mir-json`.

kmir/src/kmir/kmir.py
ehildenb: Does collapsing the tuple `(map_empty(), types,)` work?

jberthold: No, didnt. But as you are mentioning this I noticed the comma behind `types` there.
Removed that comma and the formatter accepted the single-line version :tada:  ).

kmir/src/kmir/smir.py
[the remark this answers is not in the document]
ehildenb: Can you add:

```
assert sym == 'main' or sym in fun_syms
```

Rather than this `if` block, then run the tests? If it's _just_ `main` that's missing in our current test-suite, then we should probably keep the assertion so we get notified ASAP about that assumption breaking.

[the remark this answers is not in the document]
ehildenb: If you find there are other entries missing, and it doesn't affect overall ability to build specs, then I guess it's fine to skip as is. If it's not _common_, but it still is something we should know about (because `stable-mir-json` should be extracting more info), then I would do `_LOGGER.warning(f'...')`.

ehildenb (approved):
I have one comment about changing the check to an assertion, we should probably fail as early as possible in cases we aren't sure what to do.

kmir/src/kmir/smir.py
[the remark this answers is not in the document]
jberthold: So, besides `main`, we have two `std::rt::lang_start...` functions and a `drop_in_place` (for one of them) that are never in the function table.
```python
            whitelist = [
                'main',
                'std::rt::lang_start::<()>',
                '<{closure@std::rt::lang_start<()>::{closure#0}} as std::ops::FnOnce<()>>::call_once',
                'std::ptr::drop_in_place::<{closure@std::rt::lang_start<()>::{closure#0}}>',
                '<&i32 as std::fmt::Debug>::fmt',
            ]
            assert sym in fun_syms or name in whitelist
```
The `drop` function makes me think we should keep the skipping code. Custom `drop` functions get called by the runtime (using a mechanism that we haven't investigated yet) and they can occur for many custom types in a program.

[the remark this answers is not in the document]
jberthold: We decided to leave the skipping code in place.

PR #586, opened 2025-05-15
Option to load a smir-json with prove-rs

gtrepta:
This adds `--load-smir` to `kmir prove-rs` to allow loading a smir json directly that can't be trivially generated from a single rust file.


kmir/src/kmir/__main__.py
```diff
@@ -245,6 +245,7 @@ def _arg_parser() -> ArgumentParser:
     prove_rs_parser.add_argument(
         '--save-smir', action='store_true', help='Do not delete the intermediate generated SMIR JSON file.'
     )
+    prove_rs_parser.add_argument('--load-smir', action='store_true', help='Treat the input file as a smir json.')
```
jberthold, line 248: I was thinking, why not just `--smir` ... but that's a very minor concern.

jberthold (approved):
LGTM

PR #589, opened 2025-05-16
Add struct fields when creating symbolic arg.s

jberthold:
This uses the new field types metadata for structs to create the fields symbolically when creating the call configuration.

The test program could be made to succeed by ruling out the case that fails (`a.field2 == b.field2`) as a precondition.


dkcumming (approved):
LGTM

PR #591, opened 2025-05-20
Modify `selectBlock` rules to move thunks to a side condition

jberthold:
The `#selectBlock` rules were previously matching on literal Boolean values. We want them to create branching conditions from unevaluated (`thunk`) and symbolic values.

This PR introduces a `#switchMatch` helper used as a side condition, which evaluates to the desired path conditions.
The `evaluateAsInt` wrapper around an unevaluated expression is similar to the `thunk` one but specifically expects the result to be `Integer`-valued. We could add equations for it (as a partial function) but it anyway won't get created if the expression can be evaluated.


dkcumming (approved):
LGTM

PR #592, opened 2025-05-20
Add test harness for ptoken smir json file

ehildenb:
This adds a basic dedicated testing harness for the ptoken smir harness. It:

- Uses the `smir.json` directly.
- Just executes a single entrypoint `process_transfer`.
- Does not check that the proof passes, just what the output is.


dkcumming (approved):
LGTM

PR #595, opened 2025-05-21
Enable thunking `#applyUnOp`

ehildenb:
A couple of programs were getting stuck at `#applyUnOp` for unimplemented operations, this enables thunking unary operations. In particular:

- `#applyUnOp` is put in sort `Evaluation` so that thunking can happen.
- `#compute` is renamed to `#applyBinOp`.
- The P-token test is limited to 7 iterations of the prover maximum to keep it from running too long (arbitrary limit, still takes about 10 minutes).
- Test output is updated: there are 2 tests where the rename `#compute => #applyBinOp` has an effect, and 2 where we get further into execution of programs because we thunk up `#applyUnOp` (including the P-token test).


kmir/src/tests/integration/data/prove-rs/show/symbolic-args-fail.expected
```diff
@@ -27,13 +55,38 @@
...
+   ┃  │   #selectBlock ( switchTargets ( ... branches: .Branches , otherwise: basicBlockId
+   ┃  │
+   ┃  │  (85 steps)
+   ┃  └─ 17 (vacuous, leaf)
+   ┃      #expect ( typedValue ( BoolVal ( 0 <Int size ( ARG_ARRAY7:List ) ) , ty ( 28 ) ,
+   ┃
+   ┗━━┓
+      │
+      └─ 13 (stuck, leaf)
+          #selectBlock ( switchTargets ( ... branches: branch ( 0 , basicBlockIdx ( 7 ) )
```
ehildenb, line 89: We can see we're getting `vacuous` and `stuck` nodes (which are nonsensical), and that our `#switchBlocks` is not carrying through side-conditions correctly. That will have to be investigated in a further PR.

dkcumming (approved):

PR #597, opened 2025-05-21
Resolve paths late

ehildenb:
This has been in my local tree for a while, basically don't resolve paths unless it's needed. Why not?


dkcumming (approved):
Why not?

PR #600, opened 2025-05-22
Adjust testing each start symbol for prove-rs integration tests

ehildenb:
Before, we could only `show` the output for `prove-rs` tests that had a single start symbol. This adjusts the test-harness os that we can `show` the output for multiple start symbols of these tests, and localizes the specification of which start symbols to use to the testing harness.

I'm doing this because as I was working with `symbolic-args`, I realized it's getting stuck partway through execution but we're not testing that behavior. I wanted to add it and `show` it, but I couldn't because we were already showing a different `start-symbol` for that test.

- Moves the list of start symbols to test for a given proof into the `test_integration.py` file as a list.
- Adjusts the testing logic to allow showing multiple start symbols, nad renames the expected files to include the start symbols in the name.
- Adds `main` to the list of start symbols for `symbolic-args`, and adds the expected output.
- Removes a now unused expected output file (I suspect the test was failing before and is now passing, but the expected output wasn't deleted).
- Turns off the `full_printer` for the pinocchio test, to turn down the size of the test output to something more reasonable.


dkcumming (approved):
Makes sense and looks good to me

PR #602, opened 2025-05-25
Refactor SMIRInfo Usage

ehildenb:
This PR is several refactors to how SMIRInfo is built/used to make it easier to do a later refactor I'm doing to unstick an example that was supplied. It also includes some formatting changes that were in my local work:

- Formatting:
  - The rules for `#decodeConstant` are updated to include mention of the `<k>` cell (instead of implicit).
  - Some rule spacing/newline changes.
- Places that were using `stringToken` are now using the more generic `token` from the `pyk.kast.prelude`.
- `SMIRInfo` is a bit better encapsulated, and has some more processing logic:
  - `SMIRInfo.dump` is added, which can write the original SMIR JSON output to a file.
  - A property `SMIRInfo.smir` is added, which returns the _processed_ SMIR (which has the hashes stripped off names that have hashes attached to them, ni the `items` and `functions` in particular).
  - Places where we were using `SMIRInfo._smir` are made to use `SMIRInfo.smir` now.
  - Places outside of `SMIRInfo` where we were handling the JSON directly of `stable-mir-json` are now made to always go through the `SMIRInfo` object instead. That includes in `parse/parser.py` and in `KMIR.prove_rs`.

These changes make it a bit easier to make the next logical changes I have to make, which are related to handling intrinsic to get deeper into execution.


dkcumming (approved):
Looks great

PR #604, opened 2025-05-26
Python construction of initial states

ehildenb:
This PR switches us over to fully building the initial configuration in Python instead of in K. It leaves the K construction in place, in case anyone wants to use the old kprove/krun tools to run it, but otherwise mostly ignores it.

In particular:

- Populating configuration directly in Python:
  - The cells `<functions>`, `<types>`, and `<adt-to-ty>` are populated directly with maps instead of using K functions to construct the maps in the first step.
  - `KMIR.make_call_config` takes as input a `SMIRInfo` directly, and only uses `Parser` for the small components it needs to build for making the initial configuration.
  - `KMIR.make_call_config` is enabled to make concrete configurations where the cell contents mimics that of the `configuration` declaration.
  - `KMIR.run_parsed` is renamed to `KMIR.run_smir`, and takes as input a `SMIRInfo` and uses `KMIR.make_call_config` instead of `KMIR.make_init_config` to build the initial state.
  - Unused functions `KMIR.make_init_config` and `KMIR.run_call` are removed.
- Minor testing updates:
  - Updating the expected output is sped up by skipping tests that do not have expected output when that flag is passed, and ensuring that we update expected output even if pass/fail status of test changes.
  - The `exec-smir` test-suite expected files have the state depth removed from the file name.
  - The depths for the `exec-smir` tests are adapted to the new setup (increasing depth by 1).
  - Expected output is updated (some change to `smir-exec` test output as we're calling a function directly instead now).
  - Adds two tests, one from the Kani test-suite about reading outside the bounds of a given location, and one as a way of using checked arithmetic to assert the precondition on unchecked arithmetic. Both are failing (the out of bounds read _should_ be failing, the unchecked arithmetic should not).


kmir/src/kmir/kmir.py
```diff
@@ -66,70 +67,94 @@ def kcfg_explore(self, label: str | None = None) -> Iterator[KCFGExplore]:
...
+            parse_result = parser.parse_mir_json(type, 'TypeMapping')
+            assert parse_result is not None
+            type_mapping, _ = parse_result
+            assert isinstance(type_mapping, KApply) and len(type_mapping.args) == 2
+            ty, tyinfo = type_mapping.args
+            if ty in types:
+                raise ValueError(f'Key collision in type map: {ty}')
+            types[ty] = tyinfo
+            if isinstance(tyinfo, KApply):
+                if tyinfo.label.name in ['TypeInfo::EnumType', 'TypeInfo::StructType']:
```
jberthold, line 106: `StructUnion` also has an `adt_def` field. However, we are not supporting it, other than creating the metadata and parsing it.

kmir/src/kmir/kmir.py
```diff
@@ -66,70 +67,94 @@ def kcfg_explore(self, label: str | None = None) -> Iterator[KCFGExplore]:
...
-            'ADTTOTY_CELL': KApply('mkAdtMap', (map_empty(), types)),
+            'FUNCTIONS_CELL': self._make_function_map(smir_info),
+            'TYPES_CELL': types,
+            'ADTTOTY_CELL': adts,
         }
 
-        config = self.definition.empty_config(KSort(sort))
+        _init_subst: dict[str, KInner] = {}
+        if init:
+            _subst['LOCALS_CELL'] = list_empty()
```
jberthold, line 132: If `init` is set we cannot execute functions that have arguments, maybe we can add `assert len(args_info) == 0`.

kmir/src/kmir/kmir.py
```diff
@@ -66,70 +67,94 @@ def kcfg_explore(self, label: str | None = None) -> Iterator[KCFGExplore]:
...
+        return (subst.apply(config), constraints)
 
-    def run_call(
-        self, parsed_smir: KApply, smir_json: SMIRInfo, start_symbol: str = 'main', depth: int | None = None
-    ) -> Pattern:
-        init_config, _ = self.make_call_config(parsed_smir, smir_json, start_symbol)
+    def run_smir(self, smir_info: SMIRInfo, start_symbol: str = 'main', depth: int | None = None) -> Pattern:
+        init_config, init_constraints = self.make_call_config(smir_info, start_symbol=start_symbol, init=True)
+        if len(free_vars(init_config)) > 0 or len(init_constraints) > 0:
+            raise ValueError(f'Cannot run function with variables: {start_symbol} - {free_vars(init_config)}')
```
jberthold, line 144: We _can_ run functions with arguments but only with the Haskell backend.
OK to leave as-is for now, though, anyway we would not want to keep running them with the legacy backend and will have to develop a `krun` with booster support (branch management client-side).

kmir/src/tests/integration/data/prove-rs/show/checked_arithmetic-fail.checked_add_i32.expected
```diff
@@ -0,0 +1,40 @@
...
+   ┃ constraint:
+   ┃     notBool truncate ( ARG_INT1:Int +Int ARG_INT2:Int , 32 , Signed ) ==K ARG_INT1:Int +Int ARG_INT2:Int
+   │
+   ├─ 5
+   │   #selectBlock ( switchTargets ( ... branches: branch ( 0 , basicBlockIdx ( 4 ) )
+   │
+   │  (5 steps)
+   └─ 7 (stuck, leaf)
+       #execTerminator ( terminator ( ... kind: terminatorKindCall ( ... func: operandC
+       span: 52
```
jberthold, line 37: Should be investigated, starting with a look at what is called here.

jberthold (approved):

kmir/src/tests/integration/data/prove-rs/show/checked_arithmetic-fail.checked_add_i32.expected
ehildenb: I have local work to get this one to go further into execution, I wanted to get this version committed first.

kmir/src/kmir/kmir.py
ehildenb: I guess we can make that change when we have an example for it.

kmir/src/kmir/kmir.py
```diff
@@ -66,70 +67,94 @@ def kcfg_explore(self, label: str | None = None) -> Iterator[KCFGExplore]:
...
+            elif item['mono_item_kind']['MonoItemFn']['name'] == 'main':
+                _LOGGER.warning(f'Hardcoding name "main" for item: {item_name}')
+                item_name = 'main'
+            item_ty = KApply('ty', [token(smir_info.function_symbols_reverse[item_name])])
+            parsed_item = parser.parse_mir_json(item, 'MonoItem')
+            if not parsed_item:
+                raise ValueError(f'Could not parse MonoItemKind: {parsed_item}')
+            parsed_item_kinner, _ = parsed_item
+            if type(parsed_item_kinner) is KApply and parsed_item_kinner.label.name == 'monoItemWrapper':
+                parsed_items[item_ty] = parsed_item_kinner.args[1]
```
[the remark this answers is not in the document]
ehildenb: Done.

kmir/src/kmir/kmir.py
ehildenb: Chasing this thread down turned into quite a bit more of a rabbit hole than I'm ready to do on this PR. For now, I'm going to merge as-is. I don't think it's quite correct here, because of how calling convention works, but tests seem to be passing OK.

PR #605, opened 2025-05-27
Update nix flake inputs structure and add `rv-nix-tools` to dependency update workflow

juliankuners:
This pull requests adapts the nix flake inputs structure regarding `nixpkgs` and `rv-nix-tools`. In particular, `nixpkgs` is now specified for inputs and transitive inputs by following an explicitly specified revision of `rv-nix-tools`. More thorough reasoning for this change is outlined in documentation added by this PR runtimeverification/rv-nix-tools#14. This change is done to make `nixpkgs` properly updatable by CI.

In addition, this pull requests adapts the `update-deps.yml` workflow to also update `rv-nix-tools` that is used to implicitly specify the revision of `nixpkgs`.


flake.nix
jberthold, line 1: Thanks for looking into the nix setup. 
However, this flake is currently unused, untested, and most likely broken anyway so we should probably not spend time on it until we want to un-break it anyway..

PR #613, opened 2025-06-16
Removed `Ty`s from `#applyBinOp`

dkcumming:
### Context
I tried running the sum-to-n example we had in the `run-rs/functions/` test dir, and the proof was failing as it was `thunk`ing on `#applyBinOp` with two `typedLocals` that did not have the same type as one was `TyUnknown`. The `TyUnknown` was originating from the comparison value in the guard of the loop body as it was being moved and the local was updated to `Moved` losing the `Ty` information (as `typedLocal`s get their `Ty` from the local slot they are written to).

Here is a quick breakdown of the steps where you can see this if anyone would like to run the breaking program again:

<details>

```
// sum-to-n call node(21)

node(22) Locals[5] <- ListItem ( newLocal ( ty ( 25 ) , mutabilityMut ) ) // from #setUpCalleeData

node(55) Locals[5] <- ListItem ( typedValue ( Integer ( 10, 64, false ) , ty ( 25 ) , mutabilityMut ) ) // prior to first comparison with binOpGt for loop guard

node(64) Locals[5] <- ListItem ( Moved ) // move occurs into the arg for first evaluation of binOpGt for loop guard

// binOpGt is successful as types are known, but now Locals[5] is Moved and doesn't have type info

node(188) Locals[5] <- ListItem ( typedValue ( Integer ( 9, 64, false ) , TyUnknown , mutabilityNot ) ) // prior to second comparison with binOpGt for loop guard

node(197) Locals[5] <- ListItem ( Moved ) // move occurs into the arg for second evaluation of binOpGt for loop guard

node(202) #applyBinOp ( binOpGt , typedValue ( Integer ( 9 , 64 , false ) , TyUnknown , mutabilityNot ) , typedValue ( Integer ( 0 , 64 , false ) , ty ( 25 ) , mutabilityNot ) , false )
// Note above that first arg has TyUnknown as the Ty was lost after the first move

node(203) typedValue ( thunk ( #applyBinOp ( binOpGt , typedValue ( Integer ( 9 , 64 , false ) , TyUnknown , mutabilityNot ) , typedValue ( Integer ( 0 , 64 , false ) , ty ( 25 ) , mutabilityNot ) , false ) ) , TyUnknown , mutabilityNot )
// #applyBinOp needs Ty's to be the same so it wraps it in a thunk
```

</details>

### Changes
-  Added sum-to-n-functions.rs test to prove-rs (contains loop that will trigger the problem)
- Removed `Ty` checking from `#applyBinOp` for `isArithmetic` and `isComparison` operations.

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/rt/data.md
+++ b/kmir/src/kmir/kdist/mir-semantics/rt/data.md
@@ -746,7 +746,8 @@ These operations only read a single operand.
 
 #### Arithmetic
 
-The arithmetic operations require operands of the same numeric type.
+The arithmetic operations require operands of the same numeric type, however these rules assume the `Ty`s
+are correct.
 
 | `BinOp`           |                                        | Operands can be |
 |-------------------|--------------------------------------- |-----------------|-------------------------------------- |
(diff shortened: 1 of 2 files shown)


jberthold:
> I'm OK merging this, but my impression is that we should either:
> 
> * If we can preserve type information, do so. That can be done with a `meetTy(...)` function that calculates the maximal type that is a subtype of both types (by default returns `NoType`, but if the inputs are the same returns the inputs).
> * If we're going to be removing type information to make progress at this point, then maybe we just scrap type information altogether. If every `binOp` we apply discards type information, we won't have to get very far into execution before all type information is lost.

The situation is more complex.
* This PR here is a quick-fix to remove a problem that Daniel encountered in a program with a loop (where `Moved` is written)
* The root cause is that we lose type information for a _TypedLocal_ when we write `Moved` to it. This should be corrected in a future PR.
* In the long term, we should revise our `BinOp` and `UnOp` operations to return `Value`s. We still need the operations to know the types they operate on. OTOH, the information is present in the constructor for `Integer` values. This is a long-term plan to refactor.

jberthold (approved):

PR #628, opened 2025-07-03
Assign Ty IDs to all function items

### LINKED ISSUE #619: Dead-item pruning crashes for functions not in function table
Related: https://github.com/runtimeverification/stable-mir-json/issues/42

When working with library crates, many functions won't be present in the functions lookup table in SMIR JSON.
This is a known problem for `main`, which gets a hard-coded `Ty` assigned as a workaround, but for libraries, most of the provided API will as well not be present.
The dead-item elimination code works on the `Ty` IDs and fails when a function does not have one.

Possible solutions:
* Add _all_ functions present as items in the `functions` table in SMIR JSON, in `stable-mir-json`. This is the best fix but it is unclear how to determine the correct `Ty`s.
* Assign `Ty`s to _all_ functions present as items in the `SMIRInfo` data. This is a workaround to avoid the crash.

jberthold:
The function table in SMIR JSON currently only contains functions which are called from other functions within the crate. For library crates, there can be many functions in the crate that aren't iun the table, and for binary crates, `main` is usually missing.

Previously, only `main`'s symbol was added to the table.

This change assigns negative `Ty` IDs to every item which is not present in the function table.

As a result, then dead-item pruning code now works on all functions without crash.

The previous hack to expose start symbol functions was removed from the crate tests.

Fixes #619

### DIFF
--- a/kmir/src/kmir/smir.py
+++ b/kmir/src/kmir/smir.py
@@ -79,7 +79,17 @@ class SMIRInfo:
     @cached_property
     def function_symbols(self) -> dict[int, dict]:
         fnc_symbols = {ty: sym for ty, sym, *_ in self._smir['functions'] if type(ty) is int}
+        # by convention, Ty -1 is used for 'main' if it exists
         fnc_symbols[-1] = {'NormalSym': self.main_symbol}
+
+        # function items not present in the SMIR lookup table are added with negative Ty ID
+        missing = [name for name in self.items.keys() if {'NormalSym': name} not in fnc_symbols.values()]
+
+        fake_ty = -2
+        for name in missing:
+            fnc_symbols[fake_ty] = {'NormalSym': name}
+            fake_ty -= 1
+
         return fnc_symbols
 
     @cached_property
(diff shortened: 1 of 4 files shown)


kmir/src/kmir/smir.py
```diff
@@ -79,7 +79,17 @@ def function_arguments(self) -> dict[str, list[dict]]:
...
         fnc_symbols[-1] = {'NormalSym': self.main_symbol}
+
+        # function items not present in the SMIR lookup table are added with negative Ty ID
+        missing = [name for name in self.items.keys() if {'NormalSym': name} not in fnc_symbols.values()]
+
+        fake_ty = -2
+        for name in missing:
+            fnc_symbols[fake_ty] = {'NormalSym': name}
+            fake_ty -= 1
+
```
ehildenb, line 92:
```suggestion
        # by convention, Ty -1 is used for 'main' if it exists
        # function items not present in the SMIR lookup table are added with negative Ty ID
        missing = ['main'] + [name for name in self.items.keys() if {'NormalSym': name} not in fnc_symbols.values()]
        fake_ty = -1
        for name in missing:
            fnc_symbols[fake_ty] = {'NormalSym': name}
            fake_ty -= 1

```

Just a thought.

jberthold: Not sure, because then we may get `main` twice.
I was going to remove the `main == -1` special case but the view/show output would change

PR #637, opened 2025-08-01
Hotfix resolve index projection for references/pointers

jberthold:
The projections stored within a reference or pointer may include `Index(L:Local)` , indexing with an index read from a local.
When creating a reference or pointer, the context is lost, so the actual index must be read from the local and stored as a `ConstantIndex` projection.

Also adding a test for a previous change (offset correction when passing structs holding references across stack frame boundaries)


kmir/src/tests/integration/data/exec-smir/references/array_elem_ref.rs
```diff
@@ -0,0 +1,20 @@
+// reference into array
+
+fn f(x: &u32) {
+  assert!(*x == 0);
+}
+
+fn g(x: *const u32) {
+//   assert!(unsafe{*x} == 0); // deref alignment check currently failing
```
jberthold, line 8: This line creates code for the pointer alignment check which currently creates thunks and gets stuck.
The alignment check (as well as the `unOpOffset`) require simulating an address according to the referenced type, its (original) size and alignment must be respected on offsets and when `transmute`ing pointers to numbers.

image

PR #643, opened 2025-08-06
Add a working nix flake

juliankuners:


juliankuners:
@jberthold @dkcumming
This is a WIP, but I want to ask you whether you are okay with me removing stable-mir-json as a submodule?

You can get a stable-mir-json executable in your PATH by running `nix develop` in your shell or installing stable-mir-json directly on your system. In addition, you can let vscode have stable-mir-json in its path by either starting vscode in the nix develop shell or by using `direnv allow` in combination with the direnv vscode extension and then subsequently deactivating direnv with `direnv disallow` without subsequently disabling the direnv extension.

You can also create a wrapper script that builds and runs stable-mir-json in another checkout and have that wrapper script be in your PATH with the same name, similarly by adding that to you PATH with direnv. For that, you can check the [README in kontrol-node](https://github.com/runtimeverification/kontrol-node/blob/8b7578f9aaa36cdbac0ad326f4f8644e8607f9cc/README.md).

But again, I can also revert the changes and leave it as it is.

juliankuners:
I added the respective update PR configurations to https://github.com/runtimeverification/devops/pull/237 and https://github.com/runtimeverification/devops/pull/242, for `uv2nix`, `pyproject-build-systems`, and `rv-nix-tools`.

Also, with this PR we could decide to add `kmir` to `kup`. The `kmir` and `kmir.rust` builds are cached in cachix by new GitHub workflows in this PR. `kmir.rust` is a derivation that provides a rust distribution in the PATH for `kmir`, specifically the rust distribution used to build `stable-mir-json`.

Makefile
```diff
@@ -1,10 +1,12 @@
 UV     := uv --directory kmir
```
jberthold, line 1: (while at it) we should use `--project` here
```suggestion
UV     := uv --project kmir
```

ehildenb (approved):

ehildenb:
Wow, so LGTM!

Makefile
juliankuners: Done.

PR #646, opened 2025-08-08
Implement Subslice projection and RawPtr aggregate value

jberthold:
* Aggregate kind `RawPtr`  (`RValue`) takes two arguments, a pointer and metadata, and creates a new pointer.
  No test program (yet).
* Subslice (`ProjectionElem`) : operates on arrays and slices, extracts a shorter (possibly empty) slice.
  Tested using a small program that matches the tail of a slice using `let [a, b, c, rest @ ..] = input_slice`.

Both are required for the P-token proofs to progress.

Also copied the description of all projections from the docs into the actual semantics.


dkcumming (approved):
Looks good!

jberthold:
I'll add a test later. This program from @dkcumming contains the construct but uses features we don't support yet.
```rust
use std::slice::from_raw_parts;
fn main() {
    let mut arr = [0;3];
    arr[1] = 1;
    let arr2 = unsafe { from_raw_parts((&arr).as_ptr().add(1), 2) };
    assert!(arr2[0] == arr[1]);
    assert!(arr2.len() == 2);
}
```

PR #652, opened 2025-08-13
Refactor `#traverseProjection` to include stack height parameter

Stevengre:
- Updated the `#traverseProjection` function to accept an additional `Int` parameter representing the current height of the stack.
- Adjusted all related rules to accommodate the new parameter, ensuring correct context handling during projection traversals.
- Enhanced the handling of dereferencing and context management in various projection scenarios.


kmir/src/tests/integration/data/prove-rs/.gitignore
jberthold, line 1: When I checked out the branch to test it locally, I found that this file is in the way for `prove_rs` tests... the integration test expects all files in the directory to be rust code.
image

Stevengre:
Height is solved by `#localFromFrame ( StackFrame, Local, Int )`.

PR #673, opened 2025-09-04
Implement linking for `Memory` allocations

### LINKED ISSUE #631: Include and offset all allocations in linked SMIR JSON
The `allocs` field in the linked SMIR JSON is currently unused, and the linker does not apply offsets to the contained allocations.
IDs for `allocs` need to be offset, and the AST traversal needs to retrieve all uses of allocated data and offset the IDs to refer to the correct data.

tothtamas28:
Closes #631

### DIFF
--- a/kmir/src/kmir/smir.py
+++ b/kmir/src/kmir/smir.py
@@ -7,7 +7,7 @@ from collections import deque
 from dataclasses import dataclass
 from enum import Enum
 from functools import cached_property, reduce
-from typing import TYPE_CHECKING, NewType
+from typing import TYPE_CHECKING, NamedTuple, NewType
 
 if TYPE_CHECKING:
     from pathlib import Path
@@ -420,7 +420,7 @@ class Memory(GlobalAlloc):
 @dataclass
 class Allocation:
     data: bytes  # field 'bytes'
-    # provenance
+    provenance: ProvenanceMap
     align: int
     mutable: bool  # field 'mutability'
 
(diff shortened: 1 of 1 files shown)


kmir/src/kmir/smir.py
```diff
@@ -31,6 +34,10 @@ def from_file(smir_json_file: Path) -> SMIRInfo:
     def dump(self, smir_json_file: Path) -> None:
         smir_json_file.write_text(json.dumps(self._smir))
 
+    @cached_property
+    def allocs(self) -> dict[AllocId, Memory]:
+        return {AllocId(key): Memory.from_dict(value) for key, value in self._smir['allocs']}
+
```
dkcumming, line 40: There are other cases of `GlobalAllocs`, [here are the variants](https://github.com/rust-lang/rust/blob/a2545fd6fc66b4323f555223a860c451885d1d2b/compiler/stable_mir/src/mir/alloc.rs#L12-L26) on our commit of stable mir. I think we should either handle the other cases now, or at least create some meaningful error message for the other types if we feel they won't be in `Alloc` mapping atm - but the [AllocCollector collects them](https://github.com/runtimeverification/stable-mir-json/blob/8dcacda4d94f10ea102884887e56da335e4d161c/src/printer.rs#L663-L708) so I think it is fair to handle them now.

dkcumming: @jberthold have I overlooked something?

jberthold: I think we could just go ahead without the other `GlobalAllocs` cases for now (considering this `allocs` property as specialised for `Memory`, so to speak).
Would be good to analyse a corpus of SMIR JSON (made using `test-ui`) to find out how many non-`Memory` ones we actually get... but we can merge this as step 1 either way.

kmir/src/kmir/linker.py
```diff
@@ -58,8 +56,13 @@ def apply_offset(info: SMIRInfo, offset: int) -> None:
...
 
-    # TODO adjust all alloc IDs (incl. alloc provenance)
-    # TODO then adjust alloc references during item traversal
+    for alloc in dic['allocs']:
+        alloc[0] += offset  # alloc id
+        match alloc[1]:  # memory
+            case {'Memory': memory}:
+                apply_offset_provenance(memory['provenance'], offset)
+            case _:
+                raise AssertionError()
```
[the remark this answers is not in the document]
tothtamas28: Done.

jberthold (approved):
PR += LGTM

kmir/src/kmir/smir.py
```diff
@@ -369,3 +379,58 @@ def compute_closure(start: Ty, edges: dict[Ty, set[Ty]]) -> set[Ty]:
...
+    def from_dict(dct: dict[str, Any]) -> Memory:
+        return Memory(
+            allocation=Allocation.from_dict(dct['Memory']),
+        )
+
+
+@dataclass
+class Allocation:
+    data: bytes  # field 'bytes'
+    # provenance
```
jberthold, line 423: Any particular reason not to add the provenance here?

tothtamas28: Added.

PR #677, opened 2025-09-09
Model `GlobalAllocs`

tothtamas28:
This PR adds the missing cases for `SMIR.allocs`


dkcumming (approved):
I think this looks good! @jberthold do you see anything that would need changing?

PR #678, opened 2025-09-09
Qualify function names

### LINKED ISSUE #620: Function names in linked SMIR JSON should be qualified with crate name
To avoid name collisions between functions coming from different crates in the linked SMIR JSON, the (human-readable) function names should be qualified with the crate name.
* This name change only affects how users specify `--start-symbol`, all internal operations in the semantics use the fully-qualified names. 
* Unless a specific "main crate" is identified to the linker, users will have to _always_ qualify the function names they use as start symbol with the crate name.

tothtamas28:
Closes #620


tothtamas28:
Tests are failing as expected, but before proceeding to fix them, I'd like to double check that this is the right change. @jberthold, can you take a look?

kmir/src/kmir/linker.py
```diff
@@ -42,6 +44,26 @@ def id_range(smir: SMIRInfo) -> int:
...
+        match item:
+            case {
+                'symbol_name': symbol_name,
+                'mono_item_kind': {
+                    'MonoItemFn': {
+                        'name': name,
+                    } as dct,
+                },
+            }:
+                if not '::' in name:
```
jberthold, line 61: Unfortunately, this is not a complete characterisation of local-crate functions. If the function resides in a module within the crate, the name will be prefixed with `modulename::`. So some items with `::` are nevertheless from the local crate (others aren't).
Example: `crate-tests/single-lib/main-crate` has a `testing` module, and we get: 
```
mir-semantics/kmir/src/tests/integration/data/crate-tests$ jq -c ' .items[] | [ .symbol_name, .mono_item_kind.MonoItemFn.name]' single-lib/main-crate/target/debug/linked.smir.json 
["_ZN14small_test_lib3add17h67a713ff87afe55cE","add"]
["_ZN14small_test_lib6assume17h9c6e78f80476cd7fE","assume"]
["_ZN14small_test_lib7testing17test_add_in_range17h810360fe730dd322E","testing::test_add_in_range"]
["_ZN42_$LT$$RF$T$u20$as$u20$core..fmt..Debug$GT$3fmt17h3e83eb3d49c5b5a8E","<&u128 as std::fmt::Debug>::fmt"]
["_ZN4core3fmt3num51_$LT$impl$u20$core..fmt..Debug$u20$for$u20$u128$GT$3fmt17hfb9d82c7f008e36bE","core::fmt::num::<impl std::fmt::Debug for u128>::fmt"]
["_ZN4core3ptr29drop_in_place$LT$$RF$u128$GT$17hb88ca2b7d29bccccE","std::ptr::drop_in_place::<&u128>"]
["_ZN4core9panicking13assert_failed17hdf64d315df90cf99E","core::panicking::assert_failed::<u128, u128>"]
```
It looks like we have to consider the symbol names.

jberthold: ... and re-reading the data here, I just found that `std::ptr::drop_in_place...` is actually from the `core` crate :open_mouth:  which makes it even more difficult I guess.

tothtamas28: Good point, the reason is probably this: https://github.com/runtimeverification/rust/blob/a2545fd6fc66b4323f555223a860c451885d1d2b/library/std/src/lib.rs#L532-L533

kmir/src/tests/integration/data/crate-tests/single-bin/single_exe::main.expected
```diff
@@ -5,7 +5,6 @@
 │  (225 steps)
 ├─ 3 (terminal)
 │   #EndProgram ~> .K
-│   function: main
```
jberthold, line 8: Have you checked why we are losing this function name?
I would have expected it to be `single_exe::main` now but not disappear; maybe some rendering code has to change accordingly? (not a blocker, though).

kmir/src/tests/unit/test_linker.py
```diff
@@ -0,0 +1,119 @@
...
+    expected_qualified: str
+
+
+TEST_DATA: Final[tuple[_TestData, ...]] = (
+    _TestData(
+        symbol='_ZN14small_test_lib3add17h67a713ff87afe55cE',
+        name='add',
+        expected_demangled='small_test_lib::add::h67a713ff87afe55c',
+        expected_qualified='small_test_lib::add',
+    ),
```
jberthold, line 24: We could add a (synthetic) test using the names from the comment in `mono_item_fn_name` to document behaviour.

kmir/src/tests/integration/data/crate-tests/single-bin/single_exe::main.expected
tothtamas28: I couldn't find an obvious fix, opened an issue:
* #681

PR #682, opened 2025-09-15
Fix linker for a few edge cases

tothtamas28:


jberthold (approved):
All changes make sense. I should have thought about the `->` case before...
FYI about the `bytez`, our SMIR JSON parser currently turns the `None` elements into `\0` for better performance. Long-term plans are to evaluate all constants in the SMIR before execution.

PR #683, opened 2025-09-16
527 more decoding

jberthold:
* Added a projections field to `AllocRef` so references to inner values can be created;
* Extended the `#elemSize` function to recursively compute sizes of nested arrays
    We will probably use the size stored in the layout later but might still need some of this recursive code.
* Adjusted `#withDeref` function to change `operandMove` to `operandCopy` when adding a `Deref` projection. Otherwise a call to `raw_eq` with allocated constant argument ends up writing `Moved` to the allocated constant. 
* Added  a test which uses nested arrays, and exposed the problem with `raw_eq`


dkcumming (approved):
Makes sense! Awesome!

PR #687, opened 2025-09-19
HOTFIX avoid fallback on call terminator

jberthold:
The current call terminator rules look up the function `MonoItemKind` as a match in the function map using the map key `#tyOfCall(FUNC)`. This is not supported by booster and leads to fall-backs on every function call.

The change avoids the map match by moving the key computation and lookup into the side condition in both rules (intrinsic and non-intrinsic rule). Step count is unchanged.


dkcumming (approved):
LGTM

PR #693, opened 2025-09-23
Model `TypeMetadata::EnumType`

tothtamas28:
Blocked on:
* ~#692~


kmir/src/kmir/kdist/mir-semantics/abi.md
```diff
@@ -25,48 +24,6 @@ syntax PassMode ::= "passModeIgnore"
...
-                       | "integerLengthI128"
-syntax FloatLength ::= "floatLengthF16"
-                     | "floatLengthF32"
-                     | "floatLengthF64"
-                     | "floatLengthF128"
-
-syntax AddressSpace ::= addressSpace(Int)
-
-syntax WrappingRange ::= wrappingRange(start: MIRInt, end: MIRInt)
-
```
dkcumming, line 69: Is this getting removed because we know we won't have it appear as MIR after parsing?

dkcumming: Oh it's been moved to the other files I see. The reason things were separated into the files they were is that it mirrors the structure of the Stable/Public MIR repo in the compiler. Does shifting things into other files make things easier or is it necessary for what you were doing?

tothtamas28: Yes, it was the easy solution for the following problem:
* `TypeInfo` is defined in `ty.md`.
* `TypeInfo::EnumType` depends on sorts defined in `abi.md`.
* `abi.md` depends on `ty.md`.

The alternative would have been to extract `TypeInfo` into a third module, and import `abi.md`.

tothtamas28: Opened an issue:
* #695

PR #702, opened 2025-09-25
Update from master

jberthold:
Adds type metadata changes and p-token release workflow

The workflow file must be present on the branch at the time of the push to trigger the workflow correctly


jberthold:
**NOTE** This PR should not be squash-merged but use a merge commit to clarify the status and history of the feature branch.

Stevengre (approved):

PR #710, opened 2025-09-30
Update `feature/p-token` with `master`

dkcumming:
* Enum decoding for enums without fields
* binOpOffset with zero


dkcumming:
@jberthold is this not how we merge? I guess this is not possible to do from github.com when `master -> not-master`

jberthold:
> @jberthold is this not how we merge? I guess this is not possible to do from github.com when `master -> not-master`.
> 
> Or do we select bypass the rules?

No need to bypass once there is an approval.
What needs to be different from normal is that we want to `Create a merge commit` not `Squash and Merge`

jberthold (approved):
:rocket:

PR #711, opened 2025-09-30
Adjust Makefile to avoid nix-formatting files that aren't checked in

jberthold:


dkcumming (approved):
Seems good, I didn't run it though!

PR #723, opened 2025-10-07
Remove `<adt-to-ty>`

tothtamas28:


jberthold (approved):
Nice find... 
I checked again: The `AdtDef` is in the `AggregateKind` node of the AST but we don't need it to construct values (would only need it for checking consistency)

PR #725, opened 2025-10-08
Intercept `sysvars::rent::Rent::get()` calls and produce a symbolic Rent sysvar

jberthold:
The [call `sysvars::rent::Rent::get()` in pinocchio](https://github.com/anza-xyz/pinocchio/blob/bde84880a6709bbda4da8767b5f0a42d9678d07c/sdk/pinocchio/src/sysvars/rent.rs#L231-L233) creates an [uninitialised `Rent` data structure and then loads it from the Solana Runtime using a solana-specific sys-call](https://github.com/anza-xyz/pinocchio/blob/bde84880a6709bbda4da8767b5f0a42d9678d07c/sdk/pinocchio/src/sysvars/mod.rs#L44-L67) (on non-Solana platforms the call is stubbed out).

This change to the P-token module intercepts the function call and produces a symbolic `Rent` sysvar.

Part of https://github.com/runtimeverification/solana-token/issues/37


kmir/src/kmir/kdist/mir-semantics/symbolic/p-token.md
```diff
@@ -666,6 +666,67 @@ NB Both `load_unchecked` and `load_mut_unchecked` are intercepted in the same wa
...
+        )
+      ~> #execBlockIdx(TARGET)
+    ...
+    </k>
+    <functions> FUNCTIONS </functions>
+    requires #tyOfCall(FUNC) in_keys(FUNCTIONS)
+     andBool isMonoItemKind(FUNCTIONS[#tyOfCall(FUNC)])
+     andBool #functionName({FUNCTIONS[#tyOfCall(FUNC)]}:>MonoItemKind) ==String "<sysvars::rent::Rent as sysvars::Sysvar>::get"
+    ensures 0 <=Int ?SYS_LMP_PER_BYTEYEAR andBool ?SYS_LMP_PER_BYTEYEAR <Int 1 <<Int 64
+     andBool 0 <=Int ?SYS_BURN_PCT andBool ?SYS_BURN_PCT <Int 256
```
dkcumming, line 697: I notice no ensures for [exemption_threshold](https://github.com/anza-xyz/pinocchio/blob/bde84880a6709bbda4da8767b5f0a42d9678d07c/sdk/pinocchio/src/sysvars/rent.rs#L84), by omitting it is it just totally unrestricted? I don't think this is a problem btw just asking

dkcumming (approved):
Awesome!

kmir/src/kmir/kdist/mir-semantics/symbolic/p-token.md
jberthold: It is unrestricted because it is a float which we don't really support as a value at all.
The restrictions ensured here come from the types, `u64` and `u8`, so the threshold would be restricted to "what a float can represent"... which is ... a lot.

PR #727, opened 2025-10-08
Update P-token branch with latest master

jberthold:
* Refactored a number of Python modules
* Removed `AdtDef -> Ty` mapping from configuration
* Added option to decode static allocations in python


dkcumming (approved):
LGTM

PR #745, opened 2025-10-15
Compiling SMIR JSON to kore to avoid static data in config

jberthold:
Pervasive changes to the semantics and python client code to replace the lookup tables in the config (`memory`, `functions`, `types`) by lookup functions whose equations are generated from the `SMIRInfo` i.e., from the SMIR JSON , for each program.

Previously we had `Map`-sorted config cells `functions`, `types` and `memory`, which are now _functions_ instead. The function equations have to be added and then compiled with the LLVM backend for it to work, but the configuration is much smaller and does not have any static data any more.


kmir/src/kmir/__main__.py
```diff
@@ -61,6 +62,36 @@ def _kmir_run(opts: RunOpts) -> None:
...
+    smir_info = SMIRInfo.from_file(Path(opts.file))
+
+    # with tempfile.TemporaryDirectory() as work_dir:
+    #     kmir = KMIR.from_kompiled_via_kore(smir_info, symbolic=opts.haskell_backend, target_dir=work_dir)
+    #     result = kmir.run_smir(smir_info, start_symbol=opts.start_symbol, depth=opts.depth)
+    #     print(kmir.kore_to_pretty(result))
+
+    kmir = KMIR.from_kompiled_kore(smir_info, symbolic=opts.haskell_backend)  # leaves out-kore behind
+    result = kmir.run_smir(smir_info, start_symbol=opts.start_symbol, depth=opts.depth)
+    print(kmir.kore_to_pretty(result))
```
jberthold: This would replace the `_kmir_run` implementation but currently the `from_kompiled_kore` is only implemented for symbolic execution, and the legacy backend used by `krun` does not work as expected (the added equations seem to be ignored).

kmir/src/kmir/__main__.py
```diff
@@ -209,9 +274,12 @@ def kmir(args: Sequence[str]) -> None:
...
     match opts:
         case RunOpts():
-            _kmir_run(opts)
+            if ns.command == 'run-x':
+                _kmir_run_x(opts)
+            else:
+                _kmir_run(opts)
         case GenSpecOpts():
-            _kmir_gen_spec(opts)
+            _kmir_gen_mod(opts)
```
jberthold: These changes will be reverted, obviously.

kmir/src/kmir/kmir.py
```diff
@@ -55,6 +57,96 @@ def __init__(
...
+
+        # Process LLVM definition
+        _LOGGER.info('Writing LLVM definition file')
+        llvm_def_file = LLVM_LIB_DIR / 'definition.kore'
+        llvm_def_output = target_llvm_path / 'definition.kore'
+        _insert_rules_and_write(llvm_def_file, rules, llvm_def_output)
+
+        # Run llvm-kompile-matching and llvm-kompile for LLVM
+        # TODO use pyk to do this if possible (subprocess wrapper, maybe llvm-kompile itself?)
+        # TODO align compilation options to what we use in plugin.py
```
jberthold: There certainly is a better way to run `llvm-kompile-matching` and `llvm-kompile` ?

kmir/src/kmir/kmir.py
```diff
@@ -101,6 +193,69 @@ def functions(self, smir_info: SMIRInfo) -> dict[int, KInner]:
...
+            lhs=fun_app,
+            rhs=result_kore,
+            req=None,
+            ens=None,
+            sort=SortApp('Sort' + result_sort),
+            arg_sorts=(SortApp('Sort' + arg_sort),),
+            anti_left=None,
+            priority=50,
+            uid='fubar',
+            label='fubaz',
```
jberthold: For some reason, the rules get rendered (via `_.to_axiom().text`) _without any attributes_, i.e. no priorities, UIDs , or labels. Maybe this is the (unlikely) reason why the legacy HS backend does not accept to apply them?

kmir/src/kmir/parse/parser.py
```diff
@@ -445,15 +445,29 @@ def _parse_mir_bytes_json(self, json: JSON, prod: KProduction) -> ParseResult:
...
+                def format_printable_byte(byte_val):
+                    if byte_val is None:
+                        return '\\x00'
+                    char = chr(byte_val)
+                    if char in ESCAPE_CHARS:
+                        return ESCAPE_CHARS[char]
+                    else:
+                        return char
+
+                bytes = ''.join([format_printable_byte(i) for i in json])
```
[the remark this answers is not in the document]
tothtamas28: If `bytes` is used to construct a `KToken`, you can use `bytesToken` instead: https://github.com/runtimeverification/k/blob/a5156a20308cc25e66cb9ec64f2b9ab5f21a626e/pyk/src/pyk/kast/prelude/bytes.py#L13-L14

kmir/src/kmir/__main__.py
tothtamas28: The only way to ensure that right now is by generating K instead of KORE, and calling `kompile`.

kmir/src/kmir/kmir.py
```diff
@@ -55,6 +57,96 @@ def __init__(
         KParse.__init__(self, definition_dir)
         self.llvm_library_dir = llvm_library_dir
 
+    @staticmethod
+    def from_kompiled_kore(
+        smir_info: SMIRInfo, bug_report: Path | None = None, symbolic: bool = True, target_dir: str | None = None
+    ) -> KMIR:
```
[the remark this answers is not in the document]
jberthold: I tried this but unfortunately it would cause issues with circular dependencies between kmir.py and the new compilation module. Specifically, the `KMIR.prove_rs` method (which I made static) is loading the `SMIRInfo` from the proof when reloading, but we need to use this `SMIRInfo` to produce the modified `KMIR` object to run the proof with.

I will instead make `KMIR.run_smir` a static method with the same techniques as for `KMIR.prove-rs` (for now).

kmir/src/tests/integration/test_integration.py
```diff
@@ -335,39 +333,11 @@ def test_exec_smir(
...
-    proof_dir = tmp_path / (spec.stem + 'proofs')
-    prove_opts = ProveRawOpts(spec, proof_dir=proof_dir)
-    _kmir_prove_raw(prove_opts)
-
-    claim_labels = kmir.get_claim_index(spec).labels()
-    for label in claim_labels:
-        proof = Proof.read_proof_data(proof_dir, label)
-        assert proof.passed
+    proof = KMIR.prove_rs(prove_rs_opts)
+    assert proof.passed
```
jberthold: I have removed the single test that we had which was based on `gen-spec` and `prove-raw`, as well as switching the `test_prove_termination` suite to using `prove_rs`. This means the `gen-spec` and `prove-raw` commands are now untested. I would suggest removing them because they don't fit well with the changed execution in this PR.

Stevengre (approved):
LGTM! But I need to left some information in case I forgot the key point for this pr. The key changes in this pr:
1. new functions in `value.md` to replace the original static configuration 
2. function rule generation in `kmir.py`

kmir/src/kmir/kmir.py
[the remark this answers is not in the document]
tothtamas28: No worries, let's merge this as it is. I'll follow up with a refactoring PR and will try to tackle this issue as well.

tothtamas28 (approved):

PR #751, opened 2025-10-17
Extract SMIR kompilation into module `kmir.kompile`

tothtamas28:


kmir/src/kmir/kmir.py
```diff
@@ -62,127 +58,15 @@ def __init__(
...
-                ],
-                check=True,
-            )
-            blacklist = ['definition.kore', 'interpreter', 'dt']
-            to_copy = [file.name for file in LLVM_DEF_DIR.iterdir() if file.name not in blacklist]
-            for file in to_copy:
-                _LOGGER.info(f'Copying file {file}')
-                shutil.copy2(LLVM_DEF_DIR / file, target_llvm_path / file)
-            return KMIR(target_llvm_path, None, bug_report=bug_report)
+        from .kompile import kompile_smir
```
jberthold: This creates an import loop (however only in local scope), that's why I didn't continue with factoring out the compilation code. If that's not a problem I have no objections.

jberthold (approved):
LGTM

PR #752, opened 2025-10-17
Remove unused code

### LINKED ISSUE #748: Remove `prove-raw` and `gen-spec` commands
Change #745 removed the single test that we had which was based on `gen-spec` and `prove-raw`, as well as switching the `test_prove_termination` suite to using `prove_rs`. This means the `gen-spec` and `prove-raw` commands are now untested. 
They should be removed.

_Originally posted by @jberthold in https://github.com/runtimeverification/mir-semantics/pull/745#discussion_r2438212505_

tothtamas28:
Fixes #748


jberthold (approved):
LGTM. Always good to remove unused code, nice clean-up!

PR #755, opened 2025-10-20
Refactor call config construction

tothtamas28:
Split method `make_call_config` into `_make_concrete_call_config` and `_make_symbolic_call_config`.


jberthold (approved):
LGTM

PR #757, opened 2025-10-20
replace interim-simplification 50 by simplify-each 30

jberthold:
The `--simplify-each` option only affects booster execution while `--interim-simplification` forces a fall-back to `kore` which is more expensive.


jberthold:
I have tested this locally with a P-token proof (test_ptoken_domain_data) and saw around 17% speedup (6'42'' to 5'33'')
Removing the `simplify-each` causes memory issues in the LLVM backend and huge unevaluated terms on the `<locals>` cell and anywhere the locals are used in computations (there are a few functions that do, including `getLocal`).

Stevengre (approved):
Thank you! LGTM!

PR #765, opened 2025-10-22
Implement fuzzing for Boolean and integer parameters

tothtamas28:
* Move implementation of call config generation into `kmir.kast`
* Add method `make_call_config` to `KMIR` with parameter `mode` to select between concrete, symbolic, and a newly added random mode
* Implement random argument generation for parameters of Boolean and integer types
* Many small refactorings in call config generation


jberthold (approved):
LGTM

PR #769, opened 2025-10-23
feat(decode): runtime support for structures

Stevengre:


kmir/src/kmir/kdist/mir-semantics/rt/decoding.md
```diff
@@ -72,6 +73,95 @@ rule #decodeValue(BYTES, typeInfoArrayType(ELEMTY, noTyConst))
...
+      => Aggregate(variantIdx(0), #decodeStructFieldsWithOffsets(BYTES, TYS, #structOffsets(LAYOUT)))
+  requires #structOffsets(LAYOUT) =/=K .MachineSizes
+   andBool lengthBytes(BYTES) >=Int #neededBytesForOffsets(TYS, #structOffsets(LAYOUT))
+  [preserves-definedness]
+
+// Case 2 (no layout): use sequential decoding with no padding allowed.
+rule #decodeValue(BYTES, typeInfoStructType(_, _, TYS, noLayoutShape))
+      => Aggregate(variantIdx(0), #decodeStructFields(BYTES, TYS))
+  requires lengthBytes(BYTES) ==Int #sumElemSizes(TYS)
+  [preserves-definedness]
```
jberthold: I'd say we might not want to have this fall-back, it can produce wrong data which is worse than getting stuck.
When would we _not_ have a layout available?

kmir/src/kmir/kdist/mir-semantics/rt/decoding.md
```diff
@@ -72,6 +73,95 @@ rule #decodeValue(BYTES, typeInfoArrayType(ELEMTY, noTyConst))
...
+rule #decodeStructFields(BYTES, TY TYS)
+  => ListItem(
+       #decodeValue(
+         substrBytes(BYTES, 0, #elemSize(lookupTy(TY))),
+         lookupTy(TY)
+       )
+     )
+     #decodeStructFields(substrBytes(BYTES, #elemSize(lookupTy(TY)), lengthBytes(BYTES)), TYS)
+  requires lengthBytes(BYTES) >=Int #elemSize(lookupTy(TY))
+  [preserves-definedness]
```
jberthold: As mentioned, maybe we should rather not have this 2nd fall-back case.

kmir/src/kmir/kdist/mir-semantics/rt/decoding.md
Stevengre: Deleted!

kmir/src/kmir/kdist/mir-semantics/rt/decoding.md
```diff
@@ -72,6 +73,69 @@ rule #decodeValue(BYTES, typeInfoArrayType(ELEMTY, noTyConst))
...
+// Minimum number of input bytes required to decode all fields by the chosen offsets.
+// Uses builtin maxInt to compute max(offset + size). When offsets are absent
+// (empty list), this collapses to the exact sequential size (#sumElemSizes),
+// i.e., the sum of the field sizes.
+syntax Int ::= #neededBytesForOffsets ( Tys , MachineSizes ) [function, total]
+rule #neededBytesForOffsets(.Tys, .MachineSizes) => 0
+rule #neededBytesForOffsets(TY TYS, OFFSET OFFSETS)
+  => maxInt(#msBytes(OFFSET) +Int #elemSize(lookupTy(TY)), #neededBytesForOffsets(TYS, OFFSETS))
+rule #neededBytesForOffsets(TYS, .MachineSizes) => #sumElemSizes(TYS)
+rule #neededBytesForOffsets(.Tys, _OFFSETS) => 0 [owise]
```
jberthold: I think we should probably _require_ that the `TYS` and `OFFSETS` have the same length (at the call site), or maybe return `-1` from here when they are different length and check `0 <=Int #neededBytesForOffset` in the caller to be safe.
If we don't have offsets for all struct fields we should not try to decode anything.

(BTW It took me a while to understand why we use `maxInt` for the _minimal_ amount of bytes... but I got it now.)

kmir/src/kmir/kdist/mir-semantics/rt/decoding.md
```diff
@@ -72,6 +73,69 @@ rule #decodeValue(BYTES, typeInfoArrayType(ELEMTY, noTyConst))
...
+         lookupTy(TY)
+       )
+     )
+     #decodeStructFieldsWithOffsets(BYTES, TYS, OFFSETS)
+  requires lengthBytes(BYTES) >=Int (#msBytes(OFFSET) +Int #elemSize(lookupTy(TY)))
+  [preserves-definedness]
+
+syntax Int  ::= #sumElemSizes      ( Tys )        [function, total]
+rule #sumElemSizes(.Tys) => 0
+rule #sumElemSizes(TY TYS) => #elemSize(lookupTy(TY)) +Int #sumElemSizes(TYS)
```
jberthold: This function will become unused if `#neededBytesForOffsets` requires equal length for types and offsets list.

jberthold (approved):
Looking good now, and the examples are working!
I have two suggestions for refactoring (to let fewer exceptional cases pass through) but pre-approving.

PR #773, opened 2025-10-24
Fix `AllocRefValue`

tothtamas28:


kmir/src/kmir/decoding.py
```diff
@@ -120,18 +121,30 @@ def _decode_memory_alloc_or_unable(
...
                 if len(data) == 8:
                     # single slim pointer (assumes usize == u64)
-                    return AllocRefValue(alloc_id=alloc_id, metadata=metadata)
+                    return AllocRefValue(
+                        alloc_id=alloc_id,
+                        metadata=Metadata(
+                            size=metadata_size,
+                            pointer_offset=0,
+                            origin_size=metadata_size,
+                        ),
```
tothtamas28: I made this assumption based on the previous version of `AllocRevValue.to_kast()`, I don't know if it makes sense though.

kmir/src/kmir/decoding.py
```diff
@@ -120,18 +121,30 @@ def _decode_memory_alloc_or_unable(
...
                     # expect fat pointer
+                    actual_size = DynamicSize(int.from_bytes(data[8:16], byteorder='little', signed=False))
                     return AllocRefValue(
                         alloc_id=alloc_id,
-                        metadata=DynamicSize(int.from_bytes(data[8:16], byteorder='little', signed=False)),
+                        metadata=Metadata(
+                            size=actual_size,
+                            pointer_offset=0,
+                            origin_size=actual_size,
+                        ),
```
tothtamas28: Same here.

kmir/src/kmir/decoding.py
jberthold: I think this makes sense, when we decode we won't ever have an offset, and the `origin_size` should be the same as the `size` field IIUC.

kmir/src/kmir/value.py
```diff
@@ -116,7 +124,7 @@ def to_kast(self) -> KInner: ...
 @dataclass
 class NoMetadata(MetadataSize):
     def to_kast(self) -> KInner:
-        return KApply('noMetadata')
+        return KApply('noMetadataSize')
```
Stevengre: Thank you for changing this! I encountered a problem when running `spl-token` proof because of this.

PR #776, opened 2025-10-27
Implement fuzzing for parameters of complex types

tothtamas28:
Blocked on:
* ~#773~

### DIFF
--- a/kmir/src/tests/integration/data/run-smir-random/complex-types/test.smir.json
+++ b/kmir/src/tests/integration/data/run-smir-random/complex-types/test.smir.json
@@ -1,0 +1,17597 @@
+{
+  "name": "test",
+  "crate_id": 10639679865675414000,
+  "allocs": [
+    {
+      "alloc_id": 8,
+      "ty": 75,
+      "global_alloc": {
+        "Memory": {
+          "bytes": [
+            117,
+            110,
+            115,
+            97,
+            102,
+            101,
+            32,
+            112,
+            114,
+            101,
+            99,
+            111,
+            110,
+            100,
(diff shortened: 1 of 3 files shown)


kmir/src/tests/integration/data/run-smir-random/complex-types/final-0.expected
```diff
@@ -0,0 +1,158 @@
+<kmir>
+  <k>
+    #execIntrinsic ( IntrinsicFunction ( symbol ( "assert_inhabited" ) ) , .Operands , place (... local: local ( 23 ) , projection: .ProjectionElems ) ) ~> #continueAt ( someBasicBlockIdx ( basicBlockIdx ( 8 ) ) ) ~> .K
```
jberthold: (off-topic) Interesting that these `final-*` all get stuck on `assert_inhabited`.
Is it coming [from the `from` invocation](https://github.com/runtimeverification/mir-semantics/pull/776/files#diff-c94e3756d8e816c721188ff9a06c9ad0f729021ff0a0914f894fae0e5b5eb91eR28)? 
```rust
        let b64 = u64::from(b);
```
(which could probably be expressed as `let b64 = b as u64` instead) ?
Unfortunately we don't currently get the type instance at which `assert_inhabited` is called. We need that type to implement it properly.

jberthold: Maybe but there is a function besides `test` being called here (ID 64).
BTW I just realise that these expectation files might break if we don't check in a `smir.json` files to fix the IDs.

PR #778, opened 2025-10-28
Better discriminant types, `transmute` special cases, `onInt` attributes

jberthold:
* The `rvalueDiscriminant` operation was unconditionally returning 128-bit integers, which turns out to be wrong. We now consult the layout to return the correct bit width (but still always unsigned as this seems to be an invariant).
  Closes https://github.com/runtimeverification/solana-token/issues/41
* Two special cases of `transmute` are implemented, removing `thunk`s that `transmute` would have created before.
  - round-trip `transmute`s through a given type back to the original type are eliminated
  - for a `transmute` from an `Integer` to an `enum` without fields followed by a `discriminant`  operation, the original integer value is returned as the discriminant.
*  the (partial) `onInt` function was missing `preserves-definedness` attributes on its equation so booster did not apply it


dkcumming (approved):
I think this is good, no changes necessary from what I can see.

PR #780, opened 2025-10-28
Cut point rules and section edge command in KMIR

dkcumming:
This PR adds:
- The option `--break-on-calls` to `kmir prove` which will break the proof at the rule `#setUpCalleeData` for easy linking of proof to source code
- The command `kmir section-edge` which can be provided an existing proof and an edge (two linked nodes in the kcfg) as a tuple (e.g. `4,5`) and will proof an amount of nodes breaking the edge up. In particular I find this may be useful if we can identify when a thunk is written to `<locals>` and we can section out the edge between the node it appears and the call above it. If the distance between the nodes is `n` steps, then the max sections is `n - 1` which will take 1 step. 

### Current Problems
- `--break-on-calls` is outputing the node at `#setUpCalleeData` but is also outputting the node afterwards `#setArgsFromStack` which is doubling the amount of output - I did look but can't understand why this is happening currently, however it is also not a deal breaker for me right now - I am still able to see the nodes I want and just some extra. UPDATE: This behaviour is also the same with Kontrol so I think this is a broader thing with K pr pyk

### DIFF
--- a/kmir/src/kmir/kmir.py
+++ b/kmir/src/kmir/kmir.py
@@ -28,7 +28,6 @@ from .smir import SMIRInfo
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
-    from random import Random
     from typing import Final
 
     from pyk.cterm.show import CTermShow
@@ -106,10 +105,10 @@ class KMIR(KProve, KRun, KParse):
         *,
         start_symbol: str = 'main',
         depth: int | None = None,
-        random: Random | None = None,
+        seed: int | None = None,
     ) -> Pattern:
         smir_info = smir_info.reduce_to(start_symbol)
-        mode = RandomMode(random) if random else ConcreteMode()
+        mode = RandomMode(seed) if seed else ConcreteMode()
         init_config, _ = make_call_config(
             self.definition,
             smir_info=smir_info,
--- a/kmir/src/kmir/kast.py
+++ b/kmir/src/kmir/kast.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 import logging
 from itertools import count
+from random import Random
 from typing import TYPE_CHECKING, NamedTuple
 
-from pyk.kast.inner import KApply, KSort, KVariable, Subst, build_cons
+from pyk.kast.inner import KApply, KSequence, KSort, KVariable, Subst, build_cons
 from pyk.kast.manip import free_vars, split_config_from
 from pyk.kast.prelude.collections import list_empty, list_of
-from pyk.kast.prelude.kint import eqInt, leInt
+from pyk.kast.prelude.kint import eqInt, intToken, leInt
 from pyk.kast.prelude.ml import mlEqualsTrue
 from pyk.kast.prelude.utils import token
 
@@ -29,7 +30,6 @@ from .value import (
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Mapping, Sequence
-    from random import Random
     from typing import Any, Final
 
     from pyk.kast import KInner
@@ -56,7 +56,7 @@ class SymbolicMode(NamedTuple): ...
 
 
 class RandomMode(NamedTuple):
-    random: Random
+    seed: int
 
 
 CallConfigMode = ConcreteMode | SymbolicMode | RandomMode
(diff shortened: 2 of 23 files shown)


kmir/src/kmir/__main__.py
```diff
@@ -141,6 +142,40 @@ def _kmir_prune(opts: PruneOpts) -> None:
     proof.write_proof_data()
 
 
+def _kmir_section_edge(opts: SectionEdgeOpts) -> None:
+    # Proof dir is checked to be some in arg parsing for instructive erros
```
jberthold: Do I understand this correctly that `opts.proof_dir` is never `None` here?

```suggestion
    # Proof dir is checked to be some in arg parsing for instructive errors
    assert opts.proof_dir is not None
```

kmir/src/kmir/__main__.py
```diff
@@ -141,6 +142,40 @@ def _kmir_prune(opts: PruneOpts) -> None:
...
+
+    if not APRProof.proof_data_exists(opts.id, opts.proof_dir):
+        raise ValueError(f'Proof id {opts.id} not found in proof dir {opts.proof_dir}')
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        target_path = opts.proof_dir / opts.id if opts.proof_dir is not None else Path(tmp_dir)
+
+        print(f'target_path = {target_path}')
```
jberthold: Maybe use the logger here? `_LOGGER.info` ?

kmir/src/kmir/__main__.py
```diff
@@ -141,6 +142,40 @@ def _kmir_prune(opts: PruneOpts) -> None:
...
 
 
+def _kmir_section_edge(opts: SectionEdgeOpts) -> None:
+    # Proof dir is checked to be some in arg parsing for instructive erros
+
+    if not APRProof.proof_data_exists(opts.id, opts.proof_dir):
+        raise ValueError(f'Proof id {opts.id} not found in proof dir {opts.proof_dir}')
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        target_path = opts.proof_dir / opts.id if opts.proof_dir is not None else Path(tmp_dir)
```
jberthold: If the above assertion is correct, this `mp_dir` is redundant.

kmir/src/kmir/kdist/mir-semantics/kmir.md
```diff
@@ -324,7 +324,7 @@ An operand may be a `Reference` (the only way a function could access another fu
   syntax KItem ::= #setUpCalleeData(MonoItemKind, Operands)
 
   // reserve space for local variables and copy/move arguments from old locals into their place
-  rule <k> #setUpCalleeData(
+  rule [call]: <k> #setUpCalleeData(
               monoItemFn(_, _, someBody(body((FIRST:BasicBlock _) #as BLOCKS, NEWLOCALS, _, _, _, _))),
```
jberthold: We could instrument the `terminatorKindCall` rules (which also catches intrinsic calls) but the nice thing here is that you get the function name in the `monoItemFn` on the LHS, so your solution might be better, maybe together with a cut point on the intrinsic rule for `terminatorKindCall`.

kmir/src/kmir/kmir.py
```diff
@@ -54,6 +54,19 @@ def __init__(
...
+    ) -> list[str]:
+        cut_point_rules = []
+        if break_on_calls:
+            cut_point_rules.extend(
+                [
+                    'KMIR-CONTROL-FLOW.call',
+                ]
+            )
+        return cut_point_rules
+
```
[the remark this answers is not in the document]
jberthold: Yes, definitely better to introduce this code now so we could quickly add "--break-every-step"

jberthold (commented):
Nice! I like the `section-edge` (was doing this manually using `prune` eariler this week).

A few things in the `section-edge` code should be adjusted before we can merge.

Stevengre:
> The command kmir section-edge which can be provided an existing proof and an edge (two linked nodes in the kcfg) as a tuple (e.g. 4,5) and will proof an amount of nodes breaking the edge up. In particular I find this may be useful if we can identify when a thunk is written to <locals> and we can section out the edge between the node it appears and the call above it. If the distance between the nodes is n steps, then the max sections is n - 1 which will take 1 step.

As for `thunk`, could we add a cut rule for `thunk`? Just thinking.

BTW, I deleted `thunk` rule when developing semantics to get clearer term.

tothtamas28 (approved):

dkcumming:
@Stevengre 

>As for thunk, could we add a cut rule for thunk? Just thinking.

Yes absolutely I was planning to add this too, I will make an issue and follow up on it to add that and a rule so we can see intrinsics too like Jost said. Feel free to add other ideas to https://github.com/runtimeverification/mir-semantics/issues/789

Issue #764, opened 2025-10-22
Add exec-smir tests for offset


dkcumming:
#746 Introduces support for the `binOpOffset` and adds tests to prove_rs, however I had trouble successfully adding tests to the exec-smir so that still needs to be done

PR #784, opened 2025-10-29
refactor binopoffset with symbolic pointer offset tracking

### LINKED ISSUE #709: Implement `binOpOffset`
P-Token is `thunk`ing for many proofs when applying `binOpOffset` which comes from Rust code like `unsafe { ptr.offset(num) }` which adds an offset to a raw pointer.

This is not a trivial thing to implement as we have no concrete addresses.

aslitaser:
## overview

refactors the binopoffset implementation to use symbolic offset tracking instead of eagerly updating pointer metadata. follows the approach outlined in branch 709-binopoffset-again.

### data structures
- introduced `PtrEmulation` sort to track pointer offsets symbolically
- four variants: `ptrEmulation` (wraps metadata), `ptrOffset` (valid offset), `ptrOrigSize` (original allocation size), `InvalidOffset` (out-of-bounds)
- modified `PtrLocalValue` in python to use `PtrEmulation` instead of raw `Metadata`

### pointer operations
- refactored `binOpOffset` to create offset nodes instead of immediately checking bounds
- bounds checking now happens lazily at dereference time (matches rust's unsafe semantics)
- simplified dereference rules from 4 separate cases down to 2 unified rules

### type conversions
- implemented `#convertPtrEmul` to handle array ↔ element pointer conversions
- when casting array pointer to element pointer, creates `ptrOrigSize` to remember allocation size
- when casting back, validates bounds and creates `InvalidOffset` if out of bounds

### intrinsics
- implemented `std::intrinsics::offset` (signed offset)
- implemented `ptr::add` (unsigned, forward offset)
- implemented `ptr::sub` (unsigned, backward offset)

### tests
- added 5 new prove-rs tests covering basic offset, bounds failures, negative offsets
- added corresponding exec-smir tests with generated state files

## current status

**dereferencing logic needs refinement** - the tests are failing because `#applyPtrOffset` currently uses `PointerOffset` projection for all cases. according to the notes in 709-binopoffset-again, when dereferencing `ptrOffset(N, ptrOrigSize(_))` (pointer to single element), it should use `ConstantIndex` projection instead of `PointerOffset` which expects a `Range` value.

the current implementation works for array/slice pointers but gets stuck when dereferencing offset pointers to single elements.

## related

- closes #709 (if i fix the dereference logic)
- related to #764 (adds more offset tests)
- builds on #746 (original binopoffset implementation)

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/rt/data.md
+++ b/kmir/src/kmir/kdist/mir-semantics/rt/data.md
@@ -971,9 +971,10 @@ The `getTyOf` helper applies the projections from the `Place` to determine the `
   syntax Evaluation ::= #discriminant ( Evaluation , MaybeTy ) [strict(1)]
   // ----------------------------------------------------------------
   rule <k> #discriminant(Aggregate(IDX, _), TY:Ty)
-        => Integer(#lookupDiscriminant(lookupTy(TY), IDX), 128, false) // parameters for stored u128
+        => Integer(#lookupDiscriminant(lookupTy(TY), IDX), #discriminantSize(lookupTy(TY)), false)
         ...
        </k>
+    requires #discriminantSize(lookupTy(TY)) =/=Int 0
 
   syntax Int ::= #lookupDiscriminant ( TypeInfo , VariantIdx )  [function, total]
                | #lookupDiscrAux ( Discriminants , Int ) [function]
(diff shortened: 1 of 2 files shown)


dkcumming:
Closing as progress has the been made on this and this PR is outdated currently.

PR #792, opened 2025-10-31
Bitwidth zero for discriminants

jberthold:
Since we cannot reliably determine the discriminant bit width from layout, we:
* set bitwidth to 0 for values obtained from the RValue `Discriminant`
* allow comparisons using these discriminant values on either side of comparison operations
* allow these discriminants to be used in `switchInt` terminators


dkcumming (approved):
Looks good, we might want an issue that is backlog saying to see if we can get to the bottom of what the type will really be (sounds like a big time sink of research and confusion)

jberthold:
> Looks good, we might want an issue that is backlog saying to see if we can get to the bottom of what the type will really be (sounds like a big time sink of research and confusion)

https://github.com/runtimeverification/mir-semantics/issues/793 created

PR #795, opened 2025-10-31
Set `data_len` in concrete `cheatcode_is_*` cheatcodes

tothtamas28:


tothtamas28:
> I think this makes sense, did you try it out on proofs first to see if it also worked with symbolic stuff fine? Not that I think you wouldn't but it might have been overlooked

Tested on `test_process_approve` for both symbolic and concrete, bot runs complete succesfully.

PR #800, opened 2025-11-05
Decode option data

jberthold:
Special rules for `Option<_>` (niche-encoded pointers/ref.s and normal tag encoding) which are needed to make progress in P-token.

### DIFF
--- a/kmir/src/kmir/decoding.py
+++ b/kmir/src/kmir/decoding.py
@@ -13,6 +13,7 @@ from .ty import (
     BoolT,
     EnumT,
     Initialized,
+    IntegerLength,
     IntT,
     IntTy,
     Multiple,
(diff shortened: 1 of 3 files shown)


kmir/src/kmir/decoding.py
```diff
@@ -419,5 +421,12 @@ def _extract_tag(*, data: bytes, tag_offset: MachineSize, tag: Scalar) -> tuple[
             tag_data = data[tag_offset.in_bytes : tag_offset.in_bytes + length.value]
             tag_value = int.from_bytes(tag_data, byteorder='little', signed=False)
             return tag_value, length
+        # special case: niche-encoded optional pointer, None == 0x00000000
+        case Initialized(value=Pointer(), valid_range=WrappingRange(start=1, end=0)) if (
+            data == b'\x00\x00\x00\x00\x00\x00\x00\x00'
+        ):
```
tothtamas28: Please add a decoding test for this.

PR #808, opened 2025-11-07
docs(rt): remove closure access struct note

Stevengre:


Stevengre:
Thank you! Let me close this PR

PR #813, opened 2025-11-07
fix(rt): handle zero-sized function types in decode and ZST recognition

Stevengre:
## Summary

Fix two issues preventing correct handling of zero-sized function item types (e.g., non-capturing closures):

1. **`#zeroSizedType` did not recognize `typeInfoFunType`** (`types.md`)
   — Function items are inherently zero-sized (`#elemSize` already returns 0), but `#zeroSizedType` returned `false`, preventing `rvalueRef` from materializing uninitialized closure locals.

2. **`#decodeConstant` had no rule for `typeInfoFunType`** (`data.md`)
   — Zero-sized function constants fell through to the `thunk` wrapper instead of decoding to `Aggregate(variantIdx(0), .List)`.

## Evidence

`closure-staged.rs` with `terminate-on-thunk`:

| | Steps | Result |
|---|---|---|
| Before | 44 | `thunk(#decodeConstant(constantKindZeroSized, _, typeInfoFunType(_)))` |
| After | 266 | `#EndProgram` |

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/rt/types.md
+++ b/kmir/src/kmir/kdist/mir-semantics/rt/types.md
@@ -76,29 +76,18 @@ It also implements cancellation of inverse projections (such as casting from one
 
 The `#pointeeProjection` function computes, for compatible pointee types, how to project from one pointee to the other.
 
+It uses a **source-first strategy**: always unwrap the source type (struct wrapper or array) before
+attempting to unwrap the target type. This eliminates non-deterministic overlap between source-side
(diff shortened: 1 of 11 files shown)


dkcumming (approved):
Looks good. I think it is fine to add `--terminate-on-thunk` specifically for this for now. But that it had to be added to see that we were not decoding these functions in the test suite indicates to me we should add `--terminate-on-thunk` to all the tests (something I have been thinking for some time now). This will probably create a regression but it will also show us clearly what we are not handling. That time might not be now to swap but it would be good to understand how many tests regress with this.

PR #819, opened 2025-11-09
Support for `cheatcode_is_multisig` in K

dkcumming:
Cheatcode support to symbolically instantiate a [p-interface Multisig](https://github.com/runtimeverification/solana-token/blob/fb8994fb7c50c18529db5bb73b3b6c6ad82705df/p-interface/src/state/multisig.rs#L12-L26).

```rust
/// Multisignature data.
#[repr(C)]
pub struct Multisig {
    /// Number of signers required.
    pub m: u8,

    /// Number of valid signers.
    pub n: u8,

    /// Is `true` if this structure has been initialized.
    is_initialized: u8,

    /// Signer public keys.
    pub signers: [Pubkey; MAX_SIGNERS as usize],
}
```

This cheatcode mostly duplicates the work of the `Account`, `Mint`, `Rent` cheatcodes and only differs in included content. In particular the `signers` field which is an array of pubkeys. There is a note left in the source code that the randomisation logic for fuzzing like easily create invalid entries because of the relationship of `signers` with `m` and `n`.

The proof harness `test_process_initialize_multisig` has been run and thoroughly analysed and it appears the cheatcode is working fine.

TODO (Probably not a blocker for merging though):
- Check cheatcode invocation works in solana-token repo on proofs branch for the remaining proofs

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/symbolic/p-token.md
+++ b/kmir/src/kmir/kdist/mir-semantics/symbolic/p-token.md
@@ -36,10 +36,10 @@ We model this special assumption through a special subsort of `Value` with rules
   syntax PAccount ::=
       PAccountAccount( PAcc , IAcc )    // p::Account and iface::Account structs
     | PAccountMint( PAcc , IMint )      // p::Account and iface::Mint structs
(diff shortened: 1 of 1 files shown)


jberthold (approved):
LGTM.

PR #828, opened 2025-11-10
NO OP rules for compiler hint intrinsics

### LINKED ISSUE #802: Implement more of the hint/no-op intrinsics
A number of [rustc intrinsics](https://doc.rust-lang.org/std/intrinsics/index.html#functions) represent mere hints to the compiler and are actual no-ops for the semantics.
* `cold_path` (observed, required for P-token)
* `likely`, `unlikely` (returning `bool` argument)
* `prefetch_*` (no-op)

Only `cold_path` is required at this point.
At the same time, the intrinsics implementations should be moved to their own module `intrinsics.md` (requiring an import of the `data.md` module for projection traversals, and its dependent `value.md` for values).

dkcumming:
Some instrinsics do not affect the semantics of the compiled program but are hints to the compiler. In our semantics these intrinsics are NO OPs where the call is ignored and current continuation (`<k>` cell) moves onto the next term in the sequence.

In particular there are rules implemented for these intrinsics:
- `cold_path`
- `prefetch_read_data`
- `prefetch_write_data`
- `prefetch_read_instruction`
- `prefetch_write_instruction`

By handling `cold_path` we also support `likely` and `unlikely` which are `MonoItemFn` that call the `cold_path` instrinic.

This PR also refactors the intrinsics into their only file `intrinsics.md`.

closes #802


jberthold (approved):
LGTM, and good refactoring to split it out of `kmir.md`!
BTW we also need `assert_inhabited` BTW, but implementing its idea requires type information that we don't really have.  We can make it a `NO OP`, though ([it "is not guaranteed to ever panic"](https://doc.rust-lang.org/std/intrinsics/fn.assert_inhabited.html)).

PR #836, opened 2025-11-11
add simplification to remove `#mapOffset` from symbolic key bytes lists

jberthold:
`#mapOffset` adjusts the offset parameter of references and pointers in argument lists when values containing them are read from or written to outer stack frames. This happens to the bytes of `PubKey` data in P-token, which is known to be all `Integer(_, 8, false)` but abstracted in a list-sorted variable.
Without this simplification, a path condition `KEYBYTES ==K <some key>` will not imply `#mapOffset(KEYBYTES) ==K <some key>`. This was causing vacuous branches which end up in `assert_failed` calls.
Some proofs were failing before and are passing with this simplification.


sskeirik (approved):
This looks good base on the semantics of `#mapOffset` and `#adjustRef` where `#adjustRef` is a no-op on values that do not contain references or pointers.

PR #843, opened 2025-11-14
solana-96 - create unique SysRent on first access

jberthold:
The `SysRent` was previously re-created fresh on each access (with fresh existential variables), which causes vacuous branching due to inconsistencies between the variable values.

This change creates the `SysRent` (in the unused return slot of the top-most stack frame) on first access and returns the same `SysRent` each time the accessor is called.
`initialize_mint_freeze` passes with this change + additional byte-operation simplifications (#842 )

Closes: https://github.com/runtimeverification/solana-token/issues/96

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/symbolic/p-token.md
+++ b/kmir/src/kmir/kdist/mir-semantics/symbolic/p-token.md
@@ -897,54 +897,89 @@ Only the system Rent will be stored as a value directly. The `SysRent` wrapper i
   syntax Value ::= SysRent ( PRent )
 ```
 
-```{.k .symbolic}
+Note that in case of the system-global `Rest`, it is important to always return _the same_ symbolic value.
+Therefore, the value gets created in a dedicated place on first access.
+
+```k
   rule [cheatcode-get-sys-prent]:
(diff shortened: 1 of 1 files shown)


kmir/src/kmir/kdist/mir-semantics/symbolic/p-token.md
```diff
@@ -897,54 +897,89 @@ Only the system Rent will be stored as a value directly. The `SysRent` wrapper i
...
+```
+
+If the system-global `SysRent` has already been created, it is simplify written to the destination (in a `Result` type).
+The `SysRent` is stored in the outermost stack frame's return slot (local _0), which is otherwise unused as there is no caller to return to.
+
+```k
+  rule <k> #writeSysRent(DEST) => #setLocalValue(DEST, Aggregate(variantIdx(0), ListItem(SYSRENT))) ... </k>
+       <stack>
+          STACK:List // sys-rent is stored in the outermost stack frame's return slot (local _0)
+          ListItem(StackFrame(_, _, _, _, ListItem(typedValue(SysRent(_) #as SYSRENT, _, _)) _REST))
```
Stevengre: Does it means that we don't use this return value before it finished, therefore we can use it to store whatever unchanged until the function finished? Awesome Idea!

PR #851, opened 2025-11-18
`assert_inhabited` intrinsic support

dkcumming:
The [assert_inhabited](https://doc.rust-lang.org/std/intrinsics/fn.assert_inhabited.html) intrinsic will (possibly) panic if a type in uninhabited (cannot be instantiated). E.g.
```rust
assert_inhabited<!>(); // panics (probably)
assert_inhabited<u8>(); // does nothing
```

Despite being an `assert`, panicking is not guaranteed and is dependent on the target and code generation - it is sound to perform a NO OP. Since calling the intrinsic comes from a `TerminatorKind::Call`, there must be a `BasicBlock` to go to after the call. However we observe that for `assert_inhabited<!>()` the execution continues to `noBasicBlockIdx` (which would get stuck). The semantics looks for this case and converts to `#AssertInhabitedFailure`, and otherwise performs a `NO OP`.


jberthold (approved):
LGTM.
Too bad the `iter()` invocation still does not go through....

PR #853, opened 2025-11-18
Experiment: do not llvm kompile, evaluate static data lookups in HS backend

jberthold:
Evaluates the static data lookup functions using the Haskell backend evaluator only.
The lookup functions are annotated with `no-evaluators` so they are never sent to the LLVM backend library for evaluation. Default equations are generated together with the defining equations from the SMIR JSON data.

* For `kmir prove-rs`, the prover uploads an extra module (manually) with the lookup function equations. Unfortunately, this requires us to also use the prover's `extra_module` option to ensure that the equations are in scope for the proof.

`KMIR.from_kompiled_kore(..., symbolic=True)` only appends to the _Haskell_ `definition.kore`, and uses the unmodified LLVM library. All clients of `from_kompiled_kore` will use the new HS-only evaluation automatically.
* `kmir section-edge` uses  `KMIR.from_kompiled_kore(..., symbolic=True)` 
* `kmir run --haskell-backend` uses `KMIR.from_kompiled_kore(..., symbolic=True)`
* For `kmir run` without `--haskell-backend`, we still invoke `llvm-kompile` (within `KMIR.from_kompiled_kore(..., symbolic=False)` and use the LLVM interpreter with the lookup function equations included.

### DIFF
--- a/kmir/src/kmir/kmir.py
+++ b/kmir/src/kmir/kmir.py
@@ -11,7 +11,6 @@ from pyk.cli.utils import bug_report_arg
 from pyk.cterm import CTerm, cterm_symbolic
 from pyk.kast.inner import KApply, KLabel, KSequence, KSort, KToken, KVariable, Subst
 from pyk.kast.manip import abstract_term_safely, split_config_from
-from pyk.kast.outer import KFlatModule, KImport
 from pyk.kcfg import KCFG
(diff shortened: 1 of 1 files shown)


jberthold:
Unfortunately this causes core dumps (most likely in the LLVM library) in booster during p-token proofs

PR #864, opened 2025-11-24
More byte simplification

jberthold:


jberthold:
> LGTM! But I think there should be a similar simplificiation. what if we just add some bool simplification rule for ` & mask64`?

We have masking simplifications `VAL &Int bitmask => VAL requires VAL <=Int bitmask` and also similar ones to this one but not exactly this simplification (which does not require any side condition!). I observed it in long-running proofs.

PR #870, opened 2025-11-28
fix(spl): adjustRef for SPL Token domain data

Stevengre:
- solve https://github.com/runtimeverification/solana-token/issues/115

### DIFF
--- a/kmir/src/tests/integration/test_integration.py
+++ b/kmir/src/tests/integration/test_integration.py
@@ -38,9 +38,11 @@ PROVE_RS_START_SYMBOLS = {
     'assume-cheatcode-conflict-fail': ['check_assume_conflict'],
     'transmute-bytes': ['bytes_to_u64', 'u64_to_bytes'],
     'spl_token_domain_data': [
-        'test_spl_account_is_native_branch',
         'test_spl_account_domain_data',
         'test_spl_mint_domain_data',
+        'test_spl_rent_domain_data',
+        'test_spl_account_lamports_read_then_write',
+        'test_spl_account_is_native_branch',
     ],
 }
 PROVE_RS_SHOW_SPECS = [


kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
```diff
@@ -504,6 +547,31 @@ expose the wrapped payload directly.
...
+             .Contexts
+           )
+           ~> #writeProjection(VAL)
+       ...
+      </k>
+      <stack> STACK </stack>
+    requires 0 <Int OFFSET
+     andBool OFFSET <=Int size(STACK)
+     andBool isStackFrame(STACK[OFFSET -Int 1])
+     andBool 0 <=Int I
```
jberthold: The `#writeThroughRef` only operates on `Reference` values. It seems the `WriteTo` helper sort from `data.md` could be used instead of the first `Value` argument.
Many symbol arguments have changed sort from  `Place` to `Value`, was this the reason?

kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
```diff
@@ -167,6 +167,13 @@ module KMIR-SPL-TOKEN
   rule #isSPLRentGetFunc(_) => false [owise]
   rule #isSPLRentGetFunc("Rent::get") => true   // mock harness
   rule #isSPLRentGetFunc("solana_sysvar::rent::<impl Sysvar for solana_rent::Rent>::get") => true
+
+  // Adjust references when moving across stack frames
+  rule #adjustRef(SPLRefCell(PLACE, VAL), OFFSET) => SPLRefCell(#adjustRef(PLACE, OFFSET), #adjustRef(VAL, OFFSET))
+  rule #adjustRef(SPLDataBorrow(PLACE, VAL), OFFSET) => SPLDataBorrow(#adjustRef(PLACE, OFFSET), #adjustRef(VAL, OFFSET))
+  rule #adjustRef(SPLDataBorrowMut(PLACE, VAL), OFFSET) => SPLDataBorrowMut(#adjustRef(PLACE, OFFSET), #adjustRef(VAL, OFFSET))
+  rule #adjustRef(SPLDataBuffer(VAL), OFFSET) => SPLDataBuffer(#adjustRef(VAL, OFFSET))
+  rule #adjustRef(SPLPubkeyRef(VAL), OFFSET) => SPLPubkeyRef(#adjustRef(VAL, OFFSET))
```
jberthold: The `PLACE` variables have changed sort to `Value` so they should probably not be called `PLACE` any more.
Could we implement `#adjustRef` for `SPL*` values directly instead of recurring to a `Value`? The `Place` will probably always have a known shape, but  I don't know if double references (`Deref` in the projection) can be excluded. Maybe the `Place` should become a `WrteTo`?

Stevengre: > Could we implement #adjustRef for SPL* values directly instead of recurring to a Value? 

Done!

> Maybe the Place should become a WrteTo?

`WriteTo` cannot hold the projections.

> The Place will probably always have a known shape, but I don't know if double references (Deref in the projection) can be excluded.

I don't quite understand this.

PR #871, opened 2025-12-01
Allow field(0) projection as a no-op when target value is a PtrLocal

jberthold:
The change in #811 allows pointer casts between pointers to data and pointers to "transparent wrapper structs" containing one field with the same data. While the code tries to insert `field(0, _)` projections when casting, this remediation falls short when the cast pointer is of a second degree, i.e., a pointer to a pointer.
An example of this can be observed in `Iterator<_>::next`, where a pointer `**T` is cast to a pointer `*NonNull<T>`, `NonNull` being one of these "transparent wrapper structs".

The proposed code change allows for a `field(0, _)` projection to be treated as a No-Op when applied to a non-aggregate (nor union), specifically a `PtrLocal` for the case of `next()`.

This fixes `iter.next()` for cases where `Some(element)` is returned. The case where it returns `None` is a separate issue tested but not fixed here.

### DIFF
--- a/kmir/src/tests/integration/test_integration.py
+++ b/kmir/src/tests/integration/test_integration.py
@@ -58,7 +58,7 @@ PROVE_RS_SHOW_SPECS = [
     'iterator-simple-fail',
     'unions-fail',
     'transmute-maybe-uninit-fail',
-    'iter_next_2-fail'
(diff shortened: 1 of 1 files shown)


Stevengre (approved):
LGTM!

Issue #771, opened 2025-10-24
Simplify the implementation of `binOpOffset` if possible


jberthold:
PR #746 implements `binOpOffset` by extending the `Metadata` for pointers and references.
An outline of an alternative, and possibly simpler, implementation is written [in branch `709-binOpOffset-again`](https://github.com/runtimeverification/mir-semantics/compare/709-binOpOffset-again).

* Attempt to refactor the existing implementation as described in the notes.
* Maybe combine this with #764 - adding more tests for `binOpOffset`.

PR #876, opened 2025-12-03
Special fix-up rule to apply field projection on array head elements

jberthold:
When using pointer offsets followed by dereferencing, the returned result is an array (subslice of the original).
In some cases, the value is used as a single element, for instance by projecting out a field. This PR adds a special rule for the field projection case to proceed using the head element in those cases.

Related: #771


Stevengre (approved):
LGTM!

PR #884, opened 2025-12-08
More general mint cheat code

jberthold:
* may or may not be initialized (was always uninitialized before)
* may or may not have an authorisation key (Option)
* may or may not have a freeze authorisation key (Option)

The options are now modeled by a variable in the `variantIdx` field. The payload data for `None` would be `.List` but the payload data  (`Pubkey`) is set nevertheless, assuming that it will never be read if the variant index is `0` for `None`.


sskeirik (approved):
LGTM

PR #888, opened 2025-12-10
fix(spl): avoid unreachable branch caused by `lookupDiscrAux`

Stevengre:


jberthold (approved):
Agree, this is fine to merge into the feature branch, although we want more general simplifications on master in the long term

PR #897, opened 2025-12-16
chore: merge origin master & update p/spl-token.md

Stevengre:
- #893 
- #892 
- #894 
- #896 
- [update spl/p-token.md to use #execTerminatorCall](https://github.com/runtimeverification/mir-semantics/commit/39ce2a63d6ac8edbf999882da5353b587ecb6698)

PR #898, opened 2025-12-16
fix: include function pointer arguments in call_edges for reduce_to

Stevengre:
The call_edges method now collects both:
1. Direct calls: functions used as the func operand in Call terminators
2. Indirect calls: functions passed as arguments (ZeroSized constants) that may be called via function pointers

This fixes the 'unknown function' issue when a function is passed as an argument to higher-order functions like Result::map.


jberthold (approved):

jberthold:
:+1: Good catch finding this! Maybe #897 should include this?

PR #906, opened 2026-01-13
Updates from master

tothtamas28:
Update dependency: deps/k_release (#904)


dkcumming (approved):
Nice, just remember not to squash but merge instead

PR #907, opened 2026-01-14
feat: add --to-module, --minimize-proof, and --add-module CLI options

Stevengre:
Add CLI options for semantics summarization evaluation:

1. `--to-module <file>` for `kmir show`:
   - Export proof KCFG as K module to specified file
   - Supports .json (KFlatModule JSON format) and .k (K text format)

2. `--minimize-proof` for `kmir show`:
   - Minimize the proof KCFG before displaying or exporting
   - Saves minimized proof back to disk

3. `--add-module <file>` for `kmir prove-rs`:
   - Load K module files and inject their rules into the definition
   - Can be specified multiple times for multiple modules
   - Currently supports JSON format only

These features enable the workflow:
  kmir prove-rs -> kmir show --to-module --minimize-proof -> kmir prove-rs --add-module

### DIFF
--- a/kmir/src/kmir/__main__.py
+++ b/kmir/src/kmir/__main__.py
@@ -111,7 +111,6 @@ def _write_to_module(kmir: KMIR, proof: APRProof, to_module_path: Path) -> None:
         k_module_text = kmir.pretty_print(k_module_readable)
         to_module_path.write_text(k_module_text)
     _LOGGER.info(f'Module written to: {to_module_path}')
-    print(f'Module written to: {to_module_path}')
 
 
 def _kmir_show(opts: ShowOpts) -> None:
@@ -173,6 +172,7 @@ def _kmir_show(opts: ShowOpts) -> None:
     # Handle --to-module output
     if opts.to_module:
         _write_to_module(kmir, proof, opts.to_module)
+        print(f'Module written to: {opts.to_module}')
     else:
         print('\n'.join(lines))
 


kmir/src/kmir/kompile.py
```diff
@@ -131,11 +176,12 @@ def kompile_smir(
         target_llvmdt_path.mkdir(parents=True, exist_ok=True)
         target_hs_path.mkdir(parents=True, exist_ok=True)
 
-        # Process LLVM definition
+        # Process LLVM definition (only SMIR rules, not extra module rules)
+        # Extra module rules are configuration rewrites that LLVM backend doesn't support
```
tothtamas28: This is getting a bit too complicated IMO. We should extract two helpers for the two cases of `if symbolic` (not in this PR though).

kmir/src/kmir/__main__.py
```diff
@@ -84,6 +84,36 @@ def _kmir_view(opts: ViewOpts) -> None:
...
+        def _process_sentence(sent):  # type: ignore[no-untyped-def]
+            if isinstance(sent, KRule):
+                sent = sent.let(body=remove_generated_cells(sent.body))
+            return sent
+
+        k_module_readable = k_module.let(sentences=[_process_sentence(sent) for sent in k_module.sentences])
+        k_module_text = kmir.pretty_print(k_module_readable)
+        to_module_path.write_text(k_module_text)
+    _LOGGER.info(f'Module written to: {to_module_path}')
+    print(f'Module written to: {to_module_path}')
```
tothtamas28: Consider pulling this side effect to the caller.

dkcumming (approved):
I think this is all good

PR #910, opened 2026-01-16
feat(spl): cheatcode for multisig

Stevengre:
1. Unpack/Pack function recognition - Added Multisig::unpack_from_slice and Multisig::pack_into_slice to the SPL function predicates
  2. Buffer size calculation - Added a #maybeDynamicSize rule for Multisig data buffers (355 bytes), matching the SPL Token Multisig::LEN constant
  3. New cheatcode rule - [cheatcode-is-spl-multisig] that:
    - Intercepts calls to cheatcode_is_spl_multisig
    - Creates a symbolic SPLDataBuffer with the Multisig structure:
        - m: u8 - number of signers required
      - n: u8 - number of valid signers
      - is_initialized: bool
      - signers: [Pubkey; 11] - array of 11 signer public keys
    - Initializes RefCell borrow metadata with correct buffer size (355)
    - Adds appropriate ensures clauses constraining symbolic values

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
+++ b/kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
@@ -130,6 +130,9 @@ module KMIR-SPL-TOKEN
   // spl-token rent
   rule #isSPLUnpackFunc("bincode::deserialize::<'_, solana_rent::Rent>") => true
   rule #isSPLUnpackFunc("Rent::unpack") => true
+  // spl-token multisig
+  rule #isSPLUnpackFunc("<state::Multisig as solana_program_pack::Pack>::unpack_from_slice") => true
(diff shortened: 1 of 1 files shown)


Stevengre:
Tested by:
```rust
    cheatcode_is_spl_multisig(multisig);
    let mut ms = Multisig::unpack_unchecked(&multisig.data.borrow()).unwrap();
    ms.m = 2;  // require 2 signers
    ms.n = 3;  // 3 valid signers
    ms.is_initialized = true;
    Multisig::pack(ms, &mut multisig.data.borrow_mut()).unwrap();
    let ms2 = Multisig::unpack(&multisig.data.borrow()).unwrap();
    assert!(ms2.is_initialized);
    assert_eq!(ms2.m, 2);
    assert_eq!(ms2.n, 3);
```
should be added in solana-token.

PR #916, opened 2026-01-20
Updates from master

jberthold:
* build dependency updates
* moving a simplification lemma to master
* transmute cast rule for `MaybeUninit` cases (element updates within arrays) 
* stable-mir-json upgrade


dkcumming:
The merge conflict I think is due the the changes @tothtamas28 made and I approved for the smj update. I will try to fix locally now!

PR #922, opened 2026-01-29
Add support for parallel proving

tothtamas28:


dkcumming (approved):
Tested on solana proofs. No breaking change!

PR #927, opened 2026-02-02
Added Spl support for `cheatcode_maybe_same_account`

dkcumming:
This cheatcode will retrieve two `AccountInfo`s and their corresponding `SplDataBuffer`s. If the `AccountInfo`s have the same key, then it will link all the fields between the `SplDataBuffer`s to be equal. If the keys are different then do nothing.

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
+++ b/kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
@@ -665,8 +665,8 @@ be called prior to capturing initial state and prior to executing the implementa
   // Case: keys are equal - add ensures clause to constrain SPL data equality
   // The ensures clause adds the constraint that all SPL fields must be equal
   rule <k> #maybeLinkAccounts(
-          Aggregate(variantIdx(0), ListItem(Range(KEY))),
-          Aggregate(variantIdx(0), ListItem(Range(KEY))),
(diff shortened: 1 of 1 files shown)


kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
```diff
@@ -624,6 +624,93 @@ Since float casts create thunks, we simplify this pattern directly to `PRODUCT *
...
+    [priority(30), preserves-definedness]
+
+  // Helper to evaluate keys and data, then apply constraint
+  syntax KItem ::= #maybeLinkAccounts(Evaluation, Evaluation, Evaluation, Evaluation) [seqstrict]
+
+  // Case: keys are equal - add ensures clause to constrain SPL data equality
+  // The ensures clause adds the constraint that all SPL fields must be equal
+  rule <k> #maybeLinkAccounts(
+          Aggregate(variantIdx(0), ListItem(Range(KEY))),
+          Aggregate(variantIdx(0), ListItem(Range(KEY))),
```
jberthold: Maybe better to use a `requires KEY1 ==K KEY2` here?
Although we can merge it as-is for now, because that is apparently enough to make the relevant tests pass.

jberthold (approved):
LGTM

PR #928, opened 2026-02-02
Pass `haskell_threads` to `BoosterServer`

tothtamas28:


tothtamas28:
Turns out, this is already supported in `KoreServer`, although not in the ideal way: https://github.com/runtimeverification/k/blob/fafe05e8ff3cf520c72afd4b0d156a142e6d59ed/pyk/src/pyk/kore/rpc.py#L1147-L1148

I'll open a PR upstream that prepends to `GHCRTS` rather than overwrites it.

jberthold (approved):

PR #931, opened 2026-02-03
Add cut-point rules for specific functions / intrinsics

dkcumming:
This PR enables the ability to provide strings to `kmir prove-rs` command via flag `--break-on-function <STR>` that can will then be broken on as a cut point rule if the a function call matches the string. In particular, the are added to a set in the K configuration and each function / intrinsic call is modified to follow this logic:
- If a string in `breakOnFunctions` cell is a substring of the the function / instrinsic call, then break on call (processing the terminator in the same way);
- else: continue as before 

Multiple strings can be added to the set but each needs to be provided with a new `--break-on-function` flag since I couldn' think of a separator that was common and would not appear in the the fully qualified path of a function (if we want to provide the fully qualified path).

I tested this on the local mir-semantics tests and on SPL token and both worked as expected.


kmir/src/tests/integration/data/decode-value/tmp/llvm/README.md
tothtamas28: Looks like a kompiled definition was committed by mistake.

jberthold:
As discussed today in a meeting: This seems like a very useful feature for debugging proofs and semantic corner cases.
The one drawback I see is that because the function names are held in a config cell, a proof cannot (currently) be continued from a saved KCFG with a changed set of function names, they will be loaded together with the proof.
Nevertheless this is useful, with the caveat that the proof needs to be restarted when the set of "cut functions" changes.

Maybe there is an easy way to modify the pending nodes in a proof (updating this config cell)?

dkcumming:
#960 supersedes

PR #935, opened 2026-02-04
More `BinOp::Offset` tests

dkcumming:
Adding more tests for correcting the semantics of `BinOp::Offset`

### DIFF
--- a/kmir/src/tests/integration/test_integration.py
+++ b/kmir/src/tests/integration/test_integration.py
@@ -61,9 +61,9 @@ PROVE_RS_SHOW_SPECS = [
     'transmute-maybe-uninit-fail',
     'iter_next_2-fail',
     'test_offset_from-fail',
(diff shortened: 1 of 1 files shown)


kmir/src/tests/integration/test_integration.py
```diff
@@ -61,6 +61,9 @@
     'transmute-maybe-uninit-fail',
     'iter_next_2-fail',
     'test_offset_from-fail',
+    'ref-ptr-cast-arr-offset-fail',  # Should pass
+    'ref-ptr-cast-elem-fail',  # Should pass
+    'ref-ptr-cast-elem-offset-fail',  # Should pass
```
tothtamas28: I think it's safe to remove these comments, as all the tests on this list are expected to pass and are enabled on CI.

dkcumming: I'll just remove them

PR #939, opened 2026-02-11
Update with latest master

dkcumming:
- More `BinOp::Offset` tests #935
- fix: type of offset for applyBinOp(binOpOffset, ...) #936
- Add type-correcting projections on pointer cast and related operations #937


jberthold (approved):
🚀

PR #940, opened 2026-02-13
Hotfix adjust metadata origin retention

jberthold:
* When creating a reference or pointer, set the original size to the same as the current size (was `noMetadataSize` before)
* When casting a pointer, always retain the _original_ size, not the size of the previous pointer.

This makes the `iter_next_2` test pass, i.e., the iterator's end is recognised correctly.


dkcumming (commented):
I am fairly sure this would allow UB if a pointer offsets beyond the bounds of collection. The `ORIGIN_SIZE` field is how big the collection of elements is that the pointer is inside of. Copying the `SIZE` to `ORIGIN_SIZE` will mean we are modeling the pointer in a collection bigger than what it is - thus if an offset beyond the _real_ boundary were to occur our model would not detect it if it were within the bounds of the copied `SIZE`.

Instead of this approach I did an investigation for `iter_next_2-fail.rs` to pinpoint where exactly the `ORIGIN_SIZE` was incorrectly instantiated . I found that it was a `PtrToPtr` cast that removed previously correct metadata (`dynamicSize(_)`) and replaced it with `noMetadataSize`.

I made a PR with the correction to that cast #941. Similarly with that correction (and the others from the work we did together on Friday) `InitializeMultisig` (P-Token) proof harness passes

PR #948, opened 2026-02-26
fix(rt): remove mutability guard on local variable assignment

Stevengre:
## Summary

Remove the `mutabilityOf(...) ==K mutabilityMut` guard from `#setLocalValue` in `rt/data.md`. MIR's `LocalDecl::mutability` is a source-level annotation, not an assignment constraint — the Rust compiler validates legality before emitting MIR and may reuse immutable locals across loop iterations.

- Remove mutability guard from the initialized-local `#setLocalValue` rule
- Preserve original mutability on write (`mutabilityOf(...)`) instead of forcing `mutabilityMut`
- Add regression test `immutable-local-reassign.rs` (loop variable with `mutability: Not`)

Follow-up: #949 (remove mutability tracking entirely)

## Context

A `for i in 0..2` loop variable is bound via pattern matching (`Some(i)`) on each iteration, so rustc marks it as `mutability: Not`. However, rustc reuses the same MIR local across iterations, producing repeated assignments to an immutable local. This is valid MIR — confirmed via `rustc -Z unpretty=mir` and the [`LocalDecl` documentation](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.LocalDecl.html).

Without this fix, proof execution gets stuck at step 693 on `#setLocalValue(place(local(8), .ProjectionElems), Integer(1, 64, false))` — the loop counter assignment that no rule can handle.

### Proof evidence

**Without fix (RED):**
```
APRProof: immutable-local-reassign.main
    status: ProofStatus.FAILED
    stuck: 1, failing: 1

Leaf <k>:
  #setLocalValue(place(local(8), .ProjectionElems), Integer(1, 64, false))
  function: repro
```

**With fix (GREEN):**
```
test_prove_rs[immutable-local-reassign] PASSED (53.31s)
```

## Test plan

- [x] `immutable-local-reassign.rs` passes with fix, fails (stuck) without fix
- [x] Full integration test suite (`make test-integration`)

### DIFF
--- a/.github/scripts/check-cachix-pin.sh
+++ b/.github/scripts/check-cachix-pin.sh
@@ -1,49 +1,86 @@
 #!/usr/bin/env bash
-set -xeuo pipefail
+set -euo pipefail
 
 # Kup relies on cachix registry k-framework-binary.
 CACHE="k-framework-binary"
-OWNER_REPO="$(git remote get-url origin | sed -E 's#(git@github.com:|https://github.com/)##; s#\.git$##')"
-REV="$(git rev-parse HEAD)"
-# Get the output of the nix build for kmir.
-OUT="$(nix build --no-link --json ".#kmir" | jq -r '.[0].outputs.out')"
-KEY="github:${OWNER_REPO}/${REV}#kmir"
+OWNER_REPO="${OWNER_REPO:-$(git remote get-url origin | sed -E 's#(git@github.com:|https://github.com/)##; s#\.git$##')}"
+REV="${REV:-${GITHUB_SHA:-$(git rev-parse HEAD)}}"
+UNAME_S="$(uname -s)"
+UNAME_M="$(uname -m)"
+case "${UNAME_S}-${UNAME_M}" in
+  Linux-x86_64) SYSTEM="x86_64-linux" ;;
+  Linux-aarch64 | Linux-arm64) SYSTEM="aarch64-linux" ;;
+  Darwin-x86_64) SYSTEM="x86_64-darwin" ;;
+  Darwin-arm64) SYSTEM="aarch64-darwin" ;;
+  *)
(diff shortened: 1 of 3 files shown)


dkcumming (approved):
The mutability of a `Place` is not considered when assigning at the MIR level. This is a topic that is currently in some debate amongst the rustc community.

@mariaKt made [a post in the stable-mir channel of the Rust zulip](https://rust-lang.zulipchat.com/#narrow/channel/320896-project-stable-mir/topic/Question.20about.20variable.20mutability.20in.20SMIR/with/576365656) where we were informed that mutability is ignored. They directed us to [this issue](https://github.com/rust-lang/unsafe-code-guidelines/issues/400) where they discuss if writing to an immutable `Local` should be considered UB - currently it is not.

[Miri](https://github.com/rust-lang/miri#) evaluates an assignment of a MIR `Place` by [rustc_const_eval interpreter method eval_statement](https://github.com/rust-lang/rust/blob/a2545fd6fc66b4323f555223a860c451885d1d2b/compiler/rustc_const_eval/src/interpret/step.rs#L83) which does not consider mutability at all in the evaluation.

Kani also ignores the mutability of a `Local` when [declaring them for the GOTO C backend](https://github.com/model-checking/kani/blob/0db36c39e75dd34b76e6102a460493833d22b4ae/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs#L27) as it does not at all read the [mutability field of the LocalDecl](https://github.com/rust-lang/rust/blob/a2545fd6fc66b4323f555223a860c451885d1d2b/compiler/rustc_middle/src/mir/mod.rs#L974-L977). Similarly there is no check when for the [StatementKind::Assign case](https://github.com/model-checking/kani/blob/0db36c39e75dd34b76e6102a460493833d22b4ae/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs#L91-L131) for `codegen_statement`.

PR #953, opened 2026-02-28
fix(kompile): map linked NormalSym functions to monoItemFn(noBody)

Stevengre:
## Summary

- Generate `MonoItemKind::MonoItemFn(symbol(...), defId(...), noBody)` for linked `NormalSym` entries that are present in `function_symbols` but missing from `items`.
- Keep function resolution on concrete linked symbols instead of falling back to `"** UNKNOWN FUNCTION **"` for bodyless linked functions.
- Refresh affected integration/show fixtures and normalize panic symbol hash suffixes in fixture comparison to keep snapshot output stable.
- Preserve readable `kmir show --statistics --leaves` annotations for `noBody` callees (`>> function`, `>> call span`, `>> message`) even after replacing `"** UNKNOWN FUNCTION **"` with concrete symbols.

## Context

In multi-crate SMIR, linked external functions can exist in `function_symbols` but not in local `items` (no local body). Before this change, that gap could leave `lookupFunction(Ty)` without a generated `monoItemFn` equation.

### Red vs Green

**RED (actual stuck shape in show fixtures):**
```k
#setUpCalleeData(
  monoItemFn(... name: symbol("** UNKNOWN FUNCTION **"), id: defId(38), body: noBody),
  ...
)
```
This is the unresolved-function path.

**GREEN (after fix):**
```k
#setUpCalleeData(
  monoItemFn(... name: symbol("_ZN4core9panicking5panic17h<hash>E"), id: defId(38), body: noBody),
  ...
)
```
The same path now resolves to a concrete linked symbol instead of `"** UNKNOWN FUNCTION **"`.


dkcumming:
I know I have gotten to this late, but I want to bring up something related. We are not being consistent between instrinsics and other functions. For `IntrinsicSym` we demangle the name in the `name` field of the output of `stable-mir-json`, but for `NormalSym` we use the mangled name. @ehildenb had [this PR](https://github.com/runtimeverification/stable-mir-json/pull/85) converting to using the mangled name, I am not sure why the choice was to use the mangled name over the short name was used but I am going to guess it is because it is ambiguous which `black_box` is being called.

image

^^^ Notice `black_box` and `assert_failed` ^^^

Ideally I would like use a demangled but _qualified_ name for everything. e.g.: 
-  `_ZN4core9panicking19assert_failed_inner17h1d286061ca0adfe7E` -> `core::panicking::assert_failed_inner`
- `black_box` -> [core::intrinsics::black_box](https://github.com/rust-lang/rust/blob/a2545fd6fc66b4323f555223a860c451885d1d2b/library/core/src/intrinsics/mod.rs#L3402)  or [core::hint::black_box](https://github.com/rust-lang/rust/blob/a2545fd6fc66b4323f555223a860c451885d1d2b/library/core/src/hint.rs#L388)

This would remove ambiguity and stop the need to do regex stuff to strip hashes etc.

PR #954, opened 2026-02-28
fix(decode): accept signed enum tag scalars in _extract_tag

Stevengre:
## Summary

- Relax `_extract_tag` to accept both signed and unsigned `PrimitiveInt` tag metadata (`signed=_`).
- Keep discriminant decoding based on raw tag bits (`int.from_bytes(..., signed=False)`), which is what variant dispatch needs.
- Add a decode fixture for signed-tag metadata (`enum-3-variants-0-fields-signed-tag`).
- Update the prove-rs case to a passing case and refresh its `show` snapshot.

## Problem (concrete case)

This PR is motivated by:

`kmir/src/tests/integration/data/prove-rs/transmute-u8-to-enum-changed-discriminant-signed.rs`

```rust
#[repr(i8)]
enum AccountState {
    Uninitialized = 0,
    Initialized = 1,
    Frozen = -1,
}

// main checks both byte-level and signed views of the same discriminant bits:
// 255 maps to Frozen in the u8 path
// -1 maps to Frozen in the i8 path
```

For this enum layout, the tag scalar metadata can be `signed=true` (I8), even though the stored discriminant bytes are still valid raw bytes for dispatch.

## Red (before this fix)

`_extract_tag` only matched `PrimitiveInt(..., signed=False)`. When metadata was signed, decode failed with:

```text
ValueError: Unsupported tag: Initialized(value=PrimitiveInt(length=I8, signed=True), ...)
```

That blocked the signed-discriminant path in proving.

## Green (after this fix)

- `_extract_tag` now accepts either signedness metadata.
- The same bytes are decoded as unsigned raw bits for dispatch.
- The concrete program above now proves in integration tests:
  - input case is now `transmute-u8-to-enum-changed-discriminant-signed.rs`
  - expected proof snapshot is `show/transmute-u8-to-enum-changed-discriminant-signed.main.expected`
  - terminal `#EndProgram` is reached for `main`

## Why this is the right scope

This change fixes tag extraction compatibility for signed discriminant metadata without changing enum dispatch semantics beyond that path. It is covered by both:

- decode-value fixture (`enum-3-variants-0-fields-signed-tag`) and
- end-to-end prove-rs integration snapshot for the concrete Rust program above.


kmir/src/kmir/decoding.py
```diff
@@ -445,10 +445,12 @@ def _extract_tag(*, data: bytes, tag_offset: MachineSize, tag: Scalar) -> tuple[
...
                 length=length,
-                signed=False,
+                signed=_,
             ),
             valid_range=_,
         ):
+            # Stable MIR enum discriminants are represented against the raw tag bits.
+            # Use unsigned decoding even when the scalar metadata marks the tag as signed.
             tag_data = data[tag_offset.in_bytes : tag_offset.in_bytes + length.value]
             tag_value = int.from_bytes(tag_data, byteorder='little', signed=False)
```
mariaKt: So the point is that we can still use the raw data and evaluate them as unsigned int, even if the metadata represent a signed tag?
What would happen with this modified version of the enum?
```
#[repr(i8)]
enum AccountState {
    Uninitialized = 0,
    Initialized = 1,
    Frozen = -1,
    ToCheck = 255,
}
```

PR #955, opened 2026-02-28
fix(rt): generalize direct-tag enum decoding to any variant count and discriminant

Stevengre:
## Summary

Replace the specialized \`#decodeOptionTag01\` rule (which only handled exactly 2 variants with discriminants 0/1 and the Option-like 0+1 field pattern) with a general \`#decodeEnumDirect\` decoder that handles:

- Any number of variants (not just two)
- Any discriminant values (not just 0 and 1)
- Any unsigned integer tag width (u8, u16, etc.)
- Any number of fields per variant

New K helper functions (all total with `owise` fallbacks):
- `#decodeEnumDirectFields`: given variant index, decode its fields using per-variant offsets
- `#decodeEnumDirectTag`: read tag bytes as unsigned little-endian integer
- `#nthTys`: index into per-variant field type lists
- `#nthVariantOffsets`: extract field offsets from nth variant layout

## Context

Before this change, direct-tag enum decoding was specialized to `discriminants: 0, 1` and single-byte tag checks (modeled after `Option`). Any enum with different discriminant values or more than two variants would produce `UnableToDecode`.

## Changes

- `kmir/src/kmir/kdist/mir-semantics/rt/decoding.md`: replace `#decodeOptionTag01` with general `#decodeEnumDirect` rules
- `kmir/src/tests/integration/data/prove-rs/enum-direct-tag-decode.rs`: add Rust test covering repr(u8) 0/1, repr(u8) 2/5, repr(u16) 0/256, and 3-variant enums
- `kmir/src/tests/integration/test_integration.py`: register the new prove-rs test

## Testing

- `make build` ✅
- `uv --directory kmir run pytest kmir/src/tests/integration/test_integration.py -k 'enum-direct-tag-decode' -vv` ✅

### DIFF
--- a/kmir/src/kmir/kompile.py
+++ b/kmir/src/kmir/kompile.py
@@ -521,12 +521,24 @@ def _functions(kmir: KMIR, smir_info: SMIRInfo) -> dict[int, KInner]:
         for ty in smir_info.function_symbols_reverse[item_name]:
             functions[ty] = parsed_item_kinner.args[1]
 
-    # Add intrinsic functions
+    # Add intrinsic functions and linked normal symbols that have no local body in `items`.
+    # Normal symbols must still map to `monoItemFn(..., noBody)` instead of falling back to UNKNOWN FUNCTION.
     for ty, sym in smir_info.function_symbols.items():
-        if 'IntrinsicSym' in sym and ty not in functions:
+        if ty in functions:
+            continue
+        if 'IntrinsicSym' in sym:
             functions[ty] = KApply(
                 'IntrinsicFunction',
                 [KApply('symbol(_)_LIB_Symbol_String', [stringToken(sym['IntrinsicSym'])])],
+            )
+        elif isinstance(sym.get('NormalSym'), str):
+            functions[ty] = KApply(
+                'MonoItemKind::MonoItemFn',
+                (
+                    KApply('symbol(_)_LIB_Symbol_String', (stringToken(sym['NormalSym']),)),
+                    KApply('defId(_)_BODY_DefId_Int', (intToken(ty),)),
+                    KApply('noBody_BODY_MaybeBody', ()),
+                ),
             )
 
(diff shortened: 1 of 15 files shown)


kmir/src/kmir/kdist/mir-semantics/rt/decoding.md
```diff
@@ -338,22 +337,59 @@ In both cases we expect the tag to be in the single shared field, and the discri
...
     => UnableToDecode(BYTES, ENUM_TYPE)
     [owise]
 
+  // ---------------------------------------------------------------------------
+  // #decodeEnumDirectTag: read the tag bytes as an unsigned little-endian int
+  // ---------------------------------------------------------------------------
+  syntax Int ::= #decodeEnumDirectTag ( Bytes , IntegerLength ) [function, total]
+  rule #decodeEnumDirectTag(BYTES, TAG_WIDTH)
+    => Bytes2Int(substrBytes(BYTES, 0, #byteLength(TAG_WIDTH)), LE, Unsigned)
+    requires lengthBytes(BYTES) >=Int #byteLength(TAG_WIDTH)
```
dkcumming: This is unstable in the case that `#[repr(i8)]` is provided right? It will interpret the magnitude as an unsigned? e.g. `-1 -> 255`?

dkcumming (commented):
This is great, but I think it has unpredictable results if `#[repr(T)]` where `T` &in; {`i8`,`i16`,`i32`,`i64`,`i128`,`isize`}. Could we add those to the tests or bind `signed` to `false`? Also I might be wrong in my understanding, please let me know if I am confused and missing something.

dkcumming (approved):
Okay I was wrong, all makes sense. I forgot that `Discriminants` are stored as `u128` https://github.com/runtimeverification/mir-semantics/pull/840#issue-3619512495

PR #958, opened 2026-02-28
fix(spl-token): encode multisig signers as pubkey aggregates

Stevengre:
## Summary

- Encode multisig signer elements in `cheatcode-is-spl-multisig` as `Pubkey`-shaped aggregates (`Aggregate(variantIdx(0), ListItem(Range(...)))`) instead of bare `List` payloads.
- Add prove-rs reproducer coverage for signer indexing and multisig iterator paths.

## Context

Downstream projection expects signer elements to match the `Pubkey` newtype shape. Bare list encoding breaks that structural expectation in projection/index paths over `multisig.signers`.

### Red vs Green

**RED (previous cheatcode output shape):**
```k
ListItem(Range(
  ListItem(?SplSigner0:List)
  ...
  ListItem(?SplSigner10:List)
))
```
Signer entries are bare lists, not `Pubkey` newtype aggregates.

**GREEN (current cheatcode output shape):**
```k
ListItem(Range(
  ListItem(Aggregate(variantIdx(0), ListItem(Range(?SplSigner0:List))))
  ...
  ListItem(Aggregate(variantIdx(0), ListItem(Range(?SplSigner10:List))))
))
```
Signer entries now match the expected `Pubkey`-newtype structure used by signer indexing/projection paths.

### DIFF
--- a/kmir/src/tests/integration/data/prove-rs/show/spl-multisig-iter-eq-copied-next-fail.repro.expected
+++ b/kmir/src/tests/integration/data/prove-rs/show/spl-multisig-iter-eq-copied-next-fail.repro.expected
@@ -3,10 +3,10 @@
 │   #execTerminator ( terminator ( ... kind: terminatorKindCall ( ... func: operandC
 │   span: 0
 │
-│  (1076 steps)
+│  (1077 steps)
 └─ 3 (stuck, leaf)
-    ListItem ( thunk ( operandConstant ( constOperand ( ... span: span ( 137 ) , use
-    span: 137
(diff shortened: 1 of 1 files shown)


dkcumming (approved):
Nice, makes sense with [this](https://github.com/anza-xyz/solana-sdk/blob/b14d647bebb78ee2f3295ba16729da9607579d06/pubkey/src/lib.rs#L164) being the representation of the Pubkey in the sdk at the time

```rust
pub struct Pubkey(pub(crate) [u8; 32]);
```

PR #960, opened 2026-02-28
Add cut-point rules for specific functions / intrinsics (via definition))

dkcumming:
This PR builds upon #931 modifying the approach in response to the comments on that PR. For full context read #931 _first_.

The `kmir prove-rs` flag `--break-on-function` is implemented in this PR as a compiled definition with hooked function to retrieve the function names to match on. This is similar to the already existing pattern that compiles the static data of a KMIR configuration into the definition. This allows for functions to be provided both when creating the initial proof, and when reading from disc (triggers a recompile of llvm if different flags are provided).

I added a test to demonstrate this working on functions and intrinsics, only matching those provided. I do not have a test from reading a partial proof and adding different function names - I did test it but it seemed a bit overboard for a test just now.

I did try the method with [K shell access impure function](https://github.com/runtimeverification/k/blob/master/k-distribution/include/kframework/builtin/domains.md#shell-access), however this created branching for every function call since the result was stored in a symbolic value. I couldn't figure out how to get that working concretely (I don't think it is possible but might be wrong).

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -50,7 +50,9 @@ jobs:
           git checkout "${VERSION}"
           cd ../..
           sed -i 's!__smir_version__: Final = '"'[0-9a-f]*'"'!__smir_version__: Final = '"'${VERSION}'"'!' kmir/src/kmir/__init__.py
-          git add deps/stable-mir-json kmir/src/kmir/__init__.py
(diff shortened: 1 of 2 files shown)


jberthold (approved):
LGTM overall.
I agree this should be tested and not merged if it causes a major performance regression. Would be surprised if it was causing issues, though, as the function evaluation should be fast, probably faster than getting an environment variable.

PR #967, opened 2026-03-04
fix(rt): generalize direct-tag enum decoding

Stevengre:
## Summary

- Replace specialized `#decodeOptionTag01` (2 variants, 0+1 fields, disc 0/1) with a general `#decodeEnumDirect` decoder that handles **any** number of variants, arbitrary discriminant values, and any field counts per variant
- Add comprehensive test (`enum-direct-tag-decode.rs`) covering `repr(u8)` disc 0/1, `repr(u8)` disc 2/5, `repr(u16)` disc 0/256, and 3-variant enums
- All new helper functions are `total`

Supersedes #955 — instead of adding another specialized rule for each new discriminant pattern, this uses one general decoder built on existing `#findVariantIdx` and `#decodeFieldsWithOffsets` infrastructure.

## Test plan

- [x] `make build` succeeds
- [x] `test_prove_rs[enum-direct-tag-decode]` passes (new test)
- [x] All 8 enum-related prove-rs tests pass (no regression)
- [x] All 47 decode value unit tests pass (no regression)

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/rt/decoding.md
+++ b/kmir/src/kmir/kdist/mir-semantics/rt/decoding.md
@@ -297,27 +297,24 @@ If there are no fields, the enum can be decoded by using their data as the discr
    requires TAG =/=Int DISCRIMINANT
 ```
 
-#### Enums with two variants
+#### Enums with direct tag encoding and fields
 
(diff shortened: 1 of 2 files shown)


Stevengre:
Closing in favor of pushing directly to PR #955.

PR #972, opened 2026-03-04
ci: split integration tests into parallel LLVM and Haskell jobs with shared build

### LINKED ISSUE #971: Reduce integration test CI time from 2h37m by removing redundant test executions
## Problem

Integration tests currently take **2h 37m** ([example run](https://github.com/runtimeverification/mir-semantics/actions/runs/22658250856/job/65672639933?pr=957)), which is too long for PR iteration.

### Root Cause: Redundant Test Executions

Analysis of the test suite reveals significant overlap:

| Test | Backend | Count | Wall Time | Purpose |
|---|---|---|---|---|
| `test_exec_smir[*-haskell]` | Haskell | 35 | ~70 min | Concrete execution + snapshot |
| `test_exec_smir[*-llvm]` | LLVM | 35 | ~55 min | Same, different backend |
| `test_prove_termination` | Haskell | 19 | ~25 min | Prove termination (same 19 programs as exec_smir depth=None subset) |
| `test_prove_rs` | Haskell | 49 | long tail | Symbolic verification |

**Key redundancies:**
1. **`exec_smir` runs every test twice** (LLVM + Haskell) — both verify the same snapshot output
2. **`prove_termination` tests the exact same 19 programs** already covered by `exec_smir[*-haskell]`
3. 9 out of 10 overlapping `.rs` files between `prove-rs/` and `exec-smir/` are byte-identical duplicates

### Phase 1: Remove redundant test runs (small PR, ~80 min savings)

1. **`exec_smir`: keep Haskell only, remove LLVM parametrization**
   - Haskell is the backend used for proving — bugs here have higher impact
   - Removing LLVM saves ~35 tests
2. **Remove `test_prove_termination` entirely**
   - `exec_smir[*-haskell]` already runs these 19 programs through the Haskell backend
   - `prove_rs` already covers symbolic verification

**Expected result: ~2h37m → ~1h20m**

### Phase 2: Consolidate Rust test programs (structural refactor)

- Merge `prove-rs/`, `exec-smir/`, `run-rs/` source files into a single `programs/` directory
- Remove 9 duplicated `.rs` files
- Test logic references the unified directory

### Phase 3: Declarative test configuration

- Introduce a config file (e.g., `test-config.toml`) mapping each program to its test modes (`prove`, `run-haskell`, `parse-only`)
- Single test function reads config and dispatches accordingly

## Data

134 unique Rust programs across 4 directories:
- `prove-rs/`: 84 `.rs` + 1 `.json`
- `exec-smir/`: 39 `.rs`
- `run-rs/`: 34 `.rs` (only used for smir-parse stability, not integration tests)
- `run-smir-random/`: 2 `.rs`

File overlaps: `prove-rs ∩ exec-smir` = 10 (9 identical), `prove-rs ∩ run-rs` = 13

Stevengre:
## Summary

- Extract the `kompile` build into a dedicated `build` job that pushes a pre-built Docker image to GHCR, avoiding duplicate compilation across test jobs
- Split integration tests into two parallel jobs — **LLVM** and **Haskell** — each pulling the shared image
- All tests still run; no tests are skipped or removed

## Approach

Instead of the original approach of skipping tests via pytest `-k` filters, this PR restructures the CI workflow to:

1. **Build once**: A new `build` job runs `make stable-mir-json build`, commits the Docker container state, and pushes it to `ghcr.io/runtimeverification/mir-semantics/ci:<sha>`
2. **Test in parallel**: `integration-tests-llvm` and `integration-tests-haskell` jobs pull the pre-built image and run their respective test suites (`test-integration-llvm` / `test-integration-haskell`) concurrently

This eliminates the ~40min duplicate `kompile` that previously ran before each test suite, while keeping full test coverage.

## Test plan

- [x] CI pipeline passes with the new parallel structure
- [x] Both LLVM and Haskell integration test suites complete successfully
- [x] No tests are lost compared to the previous single-job configuration

Resolves #971

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,11 +73,11 @@ jobs:
           - name: 'Haskell Proofs'
             test-args: '-k "test_prove and not test_prove_termination"'
             parallel: 6
-            timeout: 90
+            timeout: 60
           - name: 'Remainder'
             test-args: '-k "not llvm and not test_run_smir_random and not test_exec_smir and not test_prove_termination and not test_prove"'
             parallel: 6
-            timeout: 15
+            timeout: 20
     steps:
       - name: 'Check out code'
         uses: actions/checkout@v4


mariaKt:
> LLVM backend regression will be missed in the previous test, which should be handled by future test framework refactoring. But if we don't add new exec_smir test or add new exec_smir test with llvm to update expected files, the result is the same as we run CI before.
I am not sure I understand what you mean here, could you clarify?

ehildenb:
I've stratified the tests a bit. Someone shoudl check that all the expected test-suites are running somewhere, then we need to update the expected quality checks. This should run significantly faster than it was before.

PR #975, opened 2026-03-05
ci(stable-mir-ui): make manual UI workflow runnable

Stevengre:
## Summary

- make the manual `test-stable-mir-ui` workflow runnable on the self-hosted runner
- derive the Rust toolchain from repo-tracked `rust-toolchain.toml` instead of hardcoding it in the workflow
- derive the `rust-lang/rust` checkout commit from the active toolchain's `rustc -vV` output instead of maintaining a separate fixed Rust source pin
- upload the real failure outputs from inside the Docker container, including updated `skip.txt` when requested

## Context

- Stacked on #968 / `codex/import-stable-mir-ui-tests`
- The base PR adds the workflow; this PR makes that workflow usable end-to-end in the containerized setup while keeping both the installed toolchain and the checked-out Rust source tree aligned to the same nightly

## Testing

- Manual validation on `zhaoji`: rebuilt the workflow image after deriving `RUST_TOOLCHAIN=nightly-2024-11-29` from `rust-toolchain.toml`
- Manual validation on `zhaoji`: verified `make stable-mir-json` succeeds before the UI run
- Manual validation on `zhaoji`: verified `test-filter='foo or bar'` is passed through correctly
- Manual validation on `zhaoji`: verified proof artifacts and updated `skip.txt` are copied back from the container
- Local container validation: verified `rustc -vV` in the workflow image resolves `commit-hash: a2545fd6fc66b4323f555223a860c451885d1d2b` for `nightly-2024-11-29`, and the workflow's shallow `rust-lang/rust` checkout can `fetch` and `checkout` that commit successfully
- Local container validation: verified `make stable-mir-json` and `make build` succeed after aligning the Rust checkout from the active toolchain
- Local container validation: ran a single-case smoke path (`tests/ui/issues/issue-11205.rs`) through `RUST_DIR_ROOT=rust make test-stable-mir-ui`; workflow plumbing succeeded and the test reached the proof stage, while the proof itself still failed with `AssertionError: Proof did not pass`


.github/workflows/test-stable-mir-ui.yml
```diff
@@ -0,0 +1,127 @@
...
+    steps:
+      - name: 'Check out code'
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.JENKINS_GITHUB_PAT }}
+          submodules: recursive
+
+      - name: 'Resolve Rust revision from stable-mir-json submodule'
+        id: stable_mir_json
+        run: echo "rust_ref=$(git rev-parse HEAD:deps/stable-mir-json)" >> "${GITHUB_OUTPUT}"
```
mariaKt: This passed CI, so the checkout works. But can you clarify why the commit hash of `stable-mir-json` is a valid ref for `rust-lang/rust`?

mariaKt (commented):
Approved, just a question.

mariaKt (approved):
Approved, but I need a clarification.

.github/workflows/test-stable-mir-ui.yml
Stevengre: You were right to question that assumption. `deps/stable-mir-json` is a submodule of `runtimeverification/stable-mir-json`, not `rust-lang/rust`, so its gitlink is not a valid Rust source ref.

I have updated the workflow to avoid a separate hand-maintained Rust source pin in a different way: it now installs the toolchain from `rust-toolchain.toml`, reads the active toolchain commit via `rustc -vV`, and then fetches/checks out that exact `rust-lang/rust` commit for `RUST_DIR_ROOT`.

I also validated that locally in the workflow container: `rustc -vV` resolves `a2545fd6fc66b4323f555223a860c451885d1d2b` for `nightly-2024-11-29`, the shallow `rust-lang/rust` checkout can fetch and switch to it successfully, and the workflow continues through `make stable-mir-json`, `make build`, and a smoke `test-stable-mir-ui` run.

PR #980, opened 2026-03-09
fix(kmir): preserve symbolic parallel proving server selection

Stevengre:
## Summary
- Keep HS-only-for-LLVM behavior opt-in via `KMIR_HS_ONLY_SYMBOLS`:
  - parse + deduplicate env symbols in `KMIR.hs_only_symbols_from_env()`
  - build booster command args in `KMIR.kore_rpc_booster_command_from_env()`
- Only inject a custom `kore-rpc-booster` command when HS-only symbols are configured; otherwise keep the existing default RPC path.
- In symbolic `kompile_smir`, when HS-only symbols are configured:
  - skip proof-specific `llvm-kompile`
  - keep generating the proof-specific Haskell definition
  - reuse the base llvm library from kdist (`mir-semantics.llvm-library`)
- Keep parallel proving correct when no llvm library is available by selecting `KoreServer` instead of `BoosterServer`.
- Add/adjust typing imports and unit coverage for the new behavior.

## Testing
Local automated checks:
- `PYTHONPATH=kmir/src make check` (pass)
- `uv run pytest -q src/tests/unit` (pass)
- Targeted regression pytest set (including previously failing cases) passed locally:
  - `test_exec_smir[closure-call-llvm]`
  - `test_prove_rs[iter-eq-copied-take-dereftruncate]`
  - total: `6 passed in 114.45s`

Remote regression (zhaoji, accepted path):
- Command: `unset KMIR_HS_ONLY_SYMBOLS && ./run-proofs.sh -t 1800 test_process_get_account_data_size`
- Result: `ProofStatus.PASSED`, `prove_exit_code=0`, `duration_seconds=145`, `nodes=15`, `pending=0`, `failing=0`, `stuck=0`
- This confirms `kore-rpc-booster + --llvm-backend-library` still passes on the existing path without forcing HS-only mode.

GitHub checks:
- Run `22986255885` was rerun (attempt 2) after cancelling a stale long-running attempt.
- Current status at update time: all non-integration checks are green, `Integration Tests` is still pending.

### DIFF
--- a/kmir/src/kmir/kmir.py
+++ b/kmir/src/kmir/kmir.py
@@ -116,8 +116,19 @@ class KMIR(KProve, KRun, KParse):
     @contextmanager
     def kcfg_explore(self, label: str | None = None, terminate_on_thunk: bool = False) -> Iterator[KCFGExplore]:
         kore_rpc_command = KMIR.kore_rpc_booster_command_from_env()
-        if kore_rpc_command is not None:
-            _LOGGER.info(f'Passing HS-only symbols to kore-rpc-booster: {KMIR.hs_only_symbols_from_env()}')
+        if kore_rpc_command is None:
+            with cterm_symbolic(
+                self.definition,
+                self.definition_dir,
+                llvm_definition_dir=self.llvm_library_dir,
+                bug_report=self.bug_report,
+                id=label if self.bug_report is not None else None,  # NB bug report arg.s must be coherent
+                simplify_each=30,
+            ) as cts:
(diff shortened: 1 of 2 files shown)


Stevengre:
Correction with exact refs:\n\nAdded TDD-structured follow-up commits for HS-only booster wiring:\n- 0f208ed4: test(kmir): add hs-only symbol env wiring unit tests\n- 6ef9fc93: feat(kmir): pass KMIR_HS_ONLY_SYMBOLS to kore-rpc-booster\n\nRed/green evidence command:\nuv --project kmir run -- pytest kmir/src/tests/unit/test_kmir_hs_only_symbols.py -q\n- red (before wiring): 3 AttributeError failures\n- green (after wiring): 3 passed

PR #983, opened 2026-03-10
fix(kmir.md): handle projected operandMove callees

Stevengre:
## Summary

Fix call dispatch for projected `operandMove` callees.
When the callee is reached through `projectionElemDeref`, `kmir` was not computing the effective callee correctly, which left the proof stuck. This PR adds a regression for that case and updates call dispatch to use the projected place type, allowing the repro to run through to `#EndProgram`.

## Testing

Validated as a `red -> partial green -> green` series with:

```bash
make test-integration TEST_ARGS="-k spl-multisig-iter-eq-copied-next-fail"
```

Each commit in the final branch passes that scoped command on the remote validation workspace:

- `86074df5` `test(integration): add spl-multisig iter-eq copied next repro`
- `f04e1656` `fix(call): match projected operandMove callees`
- `ceb4a2a0` `fix(call): compute projected callee types through to EndProgram`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,245 +1,162 @@
 # MIR Semantics
 
-In this repository, we provide a model of the semantics of Rust's Stable MIR in K to enable symbolic execution of Rust programs and proofs of program properties.
+`mir-semantics` models the semantics of Rust Stable MIR in K. The repository also ships `kmir`, a Python CLI for building the semantics, running programs, generating specs, and inspecting proofs.
 
-Also included is the `kmir` tool, a python script that acts as a front-end to the semantics.
+For semantics details and specialized workflows, see [Further Reading](#further-reading).
 
+## Quick Start
 
-## For Developers
+### Prerequisites
 
-### KMIR Setup
-
-Pre-requisites:
-- `python >= 3.10`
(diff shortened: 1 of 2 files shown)


kmir/src/kmir/kdist/mir-semantics/kmir.md
```diff
@@ -315,11 +315,26 @@ where the returned result should go.
...
+  syntax Ty ::= #projectedCallTy(Int, ProjectionElems, List) [function, total]
+
+  rule #projectedCallTy(I, PROJS, LOCALS)
+    => {getTyOf(tyOfLocal({LOCALS[I]}:>TypedLocal), PROJS)}:>Ty
+    requires 0 <=Int I andBool I <Int size(LOCALS)
+     andBool isTypedLocal(LOCALS[I])
+     andBool isTy(getTyOf(tyOfLocal({LOCALS[I]}:>TypedLocal), PROJS))
+    [preserves-definedness]
+
+  rule #projectedCallTy(_, _, _) => ty(-1) [owise]
```
jberthold: Maybe the function could return `TyUnknown` for undesired cases, and the use site could catch that?

PR #985, opened 2026-03-13
Verify Rust Standard Library - Challenge 11: Safety of Methods for Numeric Primitive Types

dkcumming:
This PR adds infrastructure for verify-rust-std ([docs](https://model-checking.github.io/verify-rust-std/) / [github](https://github.com/model-checking/verify-rust-std/)) challenges. challenges and implements [Challenge 11: Safety of Methods for Numeric Primitives](https://model-checking.github.io/verify-rust-std/challenges/0011-floats-ints.html):

  - Proof harnesses verifying correctness under preconditions (symbolic inputs)
  - Fail harnesses verifying UB detection when preconditions are violated, with expected output
   files
  - All verify-rust-std proofs run with `--terminate-on-thunk`
  - K simplification lemmas added for shift mask identities

Part 3 (floats) is blocked: KMIR lacks float value support. Harnesses exist in to_int_unchecked-fail.rs and should be moved to to_int_unchecked.rs once passing.

Progress

Part 1 - Unsafe Integer Methods (`i8`–`i128`, `u8`–`u128`). Proofs for soundness and proofs catching UB:
- [x] `unchecked_add`
- [x] `unchecked_sub`
- [x] `unchecked_mul`
- [x] `unchecked_shl`
- [x] `unchecked_shr`
- [x] `unchecked_neg` (signed only)

Part 2 - Safe API Verification. Proofs for soundness:
- [x] `wrapping_shl` (`i8`–`i128`, `u8`–`u128`)
- [x] `wrapping_shr` (`i8`–`i128`, `u8`–`u128`)
- [x] `widening_mul` (`u8`–`u64`)
- [x] `carrying_mul` (`u8`–`u64`)

Part 3 — Float to Integer Conversion: 
- [ ] `to_int_unchecked` (`f16`, `f32`, `f64`, `f128`) See "Floats" section below

### Floats
#995 added preliminary float support for KMIR, however due to conflicts with expected behaviours with floats in the [Solana equivalance proof harness'](https://github.com/runtimeverification/solana-token/tree/proofs/specs/shared) they have since been reverted. Concrete and symbolic floats will be revisited in future work.

For contributors: The haskell backend does not contain a float module and will fail for any float hooks. This means that symbolic floats or any float expression that is not able to be rewritten by the booster backend or simplified / executed by the LLVM backend will fail. While verification of floats as mapping to real numbers remains open research, bitvector verification is well defined and would be sufficient to solve this verification problem. If A. a Float.hs module was added to the haskell-backend, and B. the z3 dependency was connected to solve queries such as this with bitvectors, then I believe this challenge is achievable.


kmir/src/tests/integration/test_integration.py
```diff
@@ -119,6 +119,240 @@ def test_prove(rs_file: Path, kmir: KMIR, update_expected_output: bool) -> None:
...
+    'unchecked_shr-fail',
+    'unchecked_neg-fail',
+    'to_int_unchecked-fail',
+]
+
+
+@pytest.mark.parametrize(
+    'rs_file',
+    VERIFY_RUST_STD_FILES,
+    ids=[spec.stem for spec in VERIFY_RUST_STD_FILES],
```
mariaKt: Minor improvement: you might use the directory name here, and combine it with the file name to produce the name of the test? That way, if there are more challenges, if there are files with the same name across different challenges the test names will still be distinct.

mariaKt (commented):
Not sure if you wanted a review, as CI failed, but overall LGTM!

mariaKt (approved):
This looks good to me, nothing new to add. Thanks for addressing the naming!

PR #1001, opened 2026-03-24
Summaries for inner_test_validate_owner

mariaKt:
### Summary

This PR adds lemma rules that summarize the execution of `expected_validate_owner_result`
and `inner_test_validate_owner` for the `validate_owner` proof. Common helpers are extracted
in `VALIDATE-OWNER-COMMON`, and intercept rules and case handlers are in dedicated
modules (`EXPECTED-VALIDATE-OWNER-RESULT-P-TOKEN-LEMMA` and
`INNER-TEST-VALIDATE-OWNER-P-TOKEN-LEMMA`). The new file includes detailed documentation
of the case analysis, module structure, and design decisions, so I chose not to repeat them here.

### Development process

Cases and their requirements were identified through proof inspection with Claude, e.g.
 examining split constraints, node functions, and branching structure to map each proof path to
a lemma rule. This provided a large amount of automation in generating the rules.

Manual effort was needed to guide the process in the right direction:
- how to make rules match properly (use of `seqstrict`, sorts, priorities, when to pattern match a constructor to identify what to match and which projections to use)
- how to reduce duplication and code size blowup from the N (registered signers) × tx_signer count dimensions
(list-based helpers, manual heating/cooling instead of per-count dispatch).

### Fail rules

The `inner_test_validate_owner` lemma includes fail rules for each error case, i.e.
rules that match when the `result` argument does not match the expected error
value. These produce a `#ValidateOwnerAssertFailed` error that causes execution
to get stuck immediately with a diagnostic message. These fail
paths were not observed in the passing proofs, I had to pass a mismatched/unexpected
Result in order to observe them. However, they are included so that if
they do occur, the proof gets stuck at the lemma level rather than needing to
proceed through the small-step semantics to eventually get stuck on a
missing `core::panicking::assert_failed_inner` body. Both outcomes are stuck,
but the lemma version is faster and produces a clearer error.

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/kmir.md
+++ b/kmir/src/kmir/kdist/mir-semantics/kmir.md
@@ -10,7 +10,7 @@ requires "intrinsics.md"
 
 requires "symbolic/p-token.md"
 requires "symbolic/spl-token.md"
-requires "symbolic/inner_test_validate_owner.md"
+// requires "symbolic/inner_test_validate_owner.md"
 ```
 
 ## Syntax of MIR in K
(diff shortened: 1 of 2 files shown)


Stevengre (approved):

mariaKt:
Let's not merge this. I tried most of the tests, and even though the savings in #steps were significant, the time saved was rather small (nothing like what I measured in test_validate_owner and test_validate_owner_multisig). I believe that this is because the branching in the current tests has already occurred. So, it is not worth having to justify the use of them.

PR #1003, opened 2026-03-26
feat(rt): stabilize pointer cast address allocation

### LINKED ISSUE #638: Extend pointer emulation to pass alignment check
(Unsafe) raw pointer dereferencing creates code for a pointer alignment check which currently creates thunks and gets stuck.
The alignment check (as well as the `unOpOffset`) require simulating an address according to the referenced type, its (original) size and alignment must be respected on offsets and when `transmute`ing pointers to numbers.

image

_Originally posted by @jberthold in https://github.com/runtimeverification/mir-semantics/pull/637#discussion_r2247117483_

Stevengre:
## Summary

Refactors the pointer-to-integer work in #1002 into a reviewable red/green history and fixes the remaining symbolic regressions in the address model.

This PR now does four things:
- adds a focused cross-frame reproducer showing what `master` cannot prove today
- introduces a stable stack allocation identity for pointer casts via `frameId`
- fixes pointer-to-integer address computation to recover layout from the actual stack slot instead of relying only on the cast source pointee type
- promotes the pointer-cast regressions that are now fully proved from `*-fail` cases into normal passing prove tests

Implementation highlights:
- adds stable frame ids to the runtime configuration and stack frames
- keys `<addressMap>` / `<exposedSet>` by stable allocation identity instead of relative stack depth
- keeps lazy address allocation for pointer-to-integer casts and `PointerExposeAddress`
- recovers byte step and allocation layout from the referenced stack slot, which fixes ZST-wrapper and element-offset alignment cases

## Commit Guide

1. `test(prove): add cross-frame pointer cast reproducer`
   - adds a red proof and expected output demonstrating the missing behavior on `master`
2. `feat(rt): stabilize stack allocation identity for ptr casts`
   - introduces stable frame ids and uses them for address allocation identity
3. `fix(rt): recover pointer cast layout from stack slots`
   - fixes byte-step / allocation-layout recovery for pointer-to-integer casts
4. `test(prove): promote resolved pointer cast regressions`
   - restores the original PR coverage as passing prove tests and removes obsolete `*-fail` expectations

## Verification

Local validation run on the rewritten branch:

```bash
uv --directory kmir run pytest src/tests/integration/test_integration.py -v --timeout=600 -k 'alignment-check or interior-mut3 or local-raw or ptr-through-wrapper or ptr-transmute-nonzero or ptr-transmute-two-locals or ptr_offset or raw-ptr-cast or ref-ptr-cast-elem-offset or ref-ptr-cast-elem or ptr-cast-cross-frame-eq'
```

Result:
- `11 passed, 266 deselected`

Additional proof checks used while validating the migrated regressions:
- `raw-ptr-cast-fail`, `ptr-through-wrapper-fail`, `local-raw-fail`, `ref-ptr-cast-elem-offset-fail`, `ref-ptr-cast-elem-fail`, and `interior-mut3-fail` now each report `ProofStatus.PASSED` with `stuck: 0` and `failing: 0`, which is why they are promoted to normal pass tests in the final commit

GitHub Actions on the rewritten history are currently in progress and should be treated as the remaining readiness gate.

## Related

- Closes #1002
- Fixes #638
- Supersedes the PTR_OFFSET / alignment-block workarounds that motivated the earlier attempts


Stevengre:
**Follow-up note from closing #1014:**

PR #1014 attempted to fix array→wrapper pointer cast projections (`[u8; 2]` → `Wrapper([u8; 2])`, nested wrappers, singleton-wrapped arrays) via `#pointeeProjection` / `#wholeArrayTargetCompatible` changes, but those rules were never reached — proofs stuck at `thunk(#cast(...))` identically with or without the changes.

This PR (#1003) resolves the simple wrapper case: `ptr-cast-array-to-wrapper-fail` is promoted to a passing test. However, two cases still carry the `-fail` suffix with modified (but not fully passing) expected output:

- **`ptr-cast-array-to-nested-wrapper-fail`**: `[u8; 2]` → `Outer(Inner([u8; 2]))` — nested transparent wrappers
- **`ptr-cast-array-to-singleton-wrapped-array-fail`**: `[u8; 2]` → `Wrapper([[u8; 2]; 1])` — singleton-array-in-wrapper

These two may need a follow-up to handle recursive wrapper/array projection in the target type during pointer casts.

PR #1005, opened 2026-03-31
fix(symbolic-spl): add sol_memset cut-point for SPLDataBuffer

Stevengre:
## Summary
- Add a symbolic cut-point for `solana_program_memory::sol_memset`
- When the target is an `SPLDataBuffer`, model the call by writing back a zeroed buffer directly

## Stack
This PR is stacked on #1007 (`sync-pr1006`). The earlier `#forceSetLocal` cleanup is already in the base branch and is not part of this diff.

## Why
`close_account` eventually calls `sol_memset(*account_data, 0, data_len)` to clear SPL account data.
Without a cut-point, the prover follows the byte-by-byte iterator path and gets stuck in `#traverseProjection` while indexing through the internal `SPLDataBuffer` aggregate.

## Change
Intercept `sol_memset` in the symbolic SPL layer and replace `SPLDataBuffer(_)` with `SPLDataBuffer(Integer(0, 8, false))` directly.

## Validation
- `make build`
- `test_process_close_account_multisig` passes with `n == 1`
  - `PASSED`
  - 355 nodes, 0 stuck, 0 pending, 90 terminal

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
+++ b/kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
@@ -719,25 +719,7 @@ zeroed representation.
       <locals> LOCALS </locals>
     </currentFrame>
     <stack> STACK => ListItem(StackFrame(OLDCALLER, OLDDEST, OLDTARGET, OLDUNWIND, LOCALS)) STACK </stack>
-    requires notBool isIntrinsicFunction(FUNC)
-     andBool notBool #functionNameMatchesEnv(getFunctionName(FUNC))
-
-  rule [spl-sol-memset-fallback-filter]:
-    <k> #execSPLSolMemset(FTY, FUNC, _BUF, _VAL, _LEN, _BUFOP, ARGS, DEST, TARGET, UNWIND, SPAN) ~> _
-      => #setUpCalleeData(FUNC, ARGS, SPAN)
-    </k>
-    <currentFunc> CALLER => FTY </currentFunc>
-    <currentFrame>
-      <currentBody> _ </currentBody>
-      <caller> OLDCALLER => CALLER </caller>
-      <dest> OLDDEST => DEST </dest>
-      <target> OLDTARGET => TARGET </target>
-      <unwind> OLDUNWIND => UNWIND </unwind>
-      <locals> LOCALS </locals>
-    </currentFrame>
-    <stack> STACK => ListItem(StackFrame(OLDCALLER, OLDDEST, OLDTARGET, OLDUNWIND, LOCALS)) STACK </stack>
-    requires notBool isIntrinsicFunction(FUNC)
-     andBool #functionNameMatchesEnv(getFunctionName(FUNC))
+    [owise]
 ```
 
 ## Rent sysvar handling
--- a/kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
+++ b/kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
@@ -107,6 +107,10 @@ module KMIR-SPL-TOKEN
   syntax Bool ::= #isSplPubkey ( List ) [function, total]
   rule #isSplPubkey(KEY) => size(KEY) ==Int 32 andBool allBytes(KEY)
 
+  syntax Bool ::= #isZeroMemsetValue ( Value ) [function, total]
+  rule #isZeroMemsetValue(Integer(0, _, _)) => true
+  rule #isZeroMemsetValue(_) => false [owise]
+
   // Construct a 32-byte pubkey List from individual Int variables.
   // When used with existential variables (?Var:Int), this produces a concrete List structure
   // that ==K can decompose element-wise, avoiding opaque symbolic List equality in SMT.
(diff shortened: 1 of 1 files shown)


kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
```diff
@@ -602,6 +606,32 @@ The `#initBorrow` helper resets borrow counters to 0 and sets the correct dynami
...
+zeroed representation.
+
+```{.k .symbolic}
+  // sol_memset(buf, val, len) - intercept when target is SPLDataBuffer
+  rule [spl-sol-memset]:
+    <k> #execTerminatorCall(_, FUNC,
+          BUF:Operand _VAL:Operand _LEN:Operand .Operands,
+          _DEST, TARGET, _UNWIND, _SPAN) ~> _CONT
+      => #splSolMemset(#withDeref(BUF), BUF) ~> #continueAt(TARGET)
+    </k>
```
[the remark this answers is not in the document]
Stevengre: Fixed in 82ee9043. I narrowed the cut-point so it only fires for the case this proof relies on: zero-filling an entire recognized SPL buffer. The rule now evaluates `val` and `len`, requires `val == 0`, and requires `len` to match the full known `SPLDataBuffer` length for the recognized Account/Mint/Rent/Multisig shapes. Non-zero fills and partial writes no longer go through the cut-point, so they are no longer modeled as full zeroization.

kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
```diff
@@ -602,6 +673,38 @@ The `#initBorrow` helper resets borrow counters to 0 and sets the correct dynami
...
+
+```{.k .symbolic}
+  // sol_memset(buf, val, len) - intercept when target is SPLDataBuffer
+  rule [spl-sol-memset]:
+    <k> #execTerminatorCall(_, FUNC,
+          BUF:Operand VAL:Operand LEN:Operand .Operands,
+          _DEST, TARGET, _UNWIND, _SPAN) ~> _CONT
+      => #splSolMemset(#withDeref(BUF), #withDeref(VAL), #withDeref(LEN), BUF) ~> #continueAt(TARGET)
+    </k>
+    requires #isSPLSolMemsetFunc(#functionName(FUNC))
```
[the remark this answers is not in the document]
Stevengre: Fixed in `e183778b`. The `sol_memset` hook still fast-paths the zero/full-buffer SPL case, but unmatched call shapes now fall back to the ordinary call semantics instead of getting stuck in the helper.

kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
```diff
@@ -254,6 +262,69 @@ module KMIR-SPL-TOKEN
...
+       [priority(30)]
+
+  rule #splBufferLen(
+         SPLDataBuffer(
+           Aggregate(variantIdx(0),
+             ListItem(Integer(_, 8, false))
+             ListItem(Integer(_, 8, false))
+             ListItem(BoolVal(_))
+             ListItem(Range(
+               ListItem(_) ListItem(_) ListItem(_)
```
mariaKt, line 319: This is a place affected by MAX_SIGNERS, but we might easily miss it (no mention of signers anywhere). Consider adding a comment next to it, e,g, // len = MAX_SIGNERS ?

kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
```diff
@@ -602,6 +673,87 @@ The `#initBorrow` helper resets borrow counters to 0 and sets the correct dynami
...
+    requires #isZeroMemsetValue(VAL)
+     andBool LEN ==Int #splBufferLen(BUF)
+     andBool 0 <Int #splBufferLen(BUF)
+
+  rule [spl-sol-memset-fallback-intrinsic]:
+    <k> #execSPLSolMemset(_, FUNC, _BUF, _VAL, _LEN, _BUFOP, ARGS, DEST, TARGET, _UNWIND, SPAN) ~> _
+      => #execIntrinsic(FUNC, ARGS, DEST, SPAN) ~> #continueAt(TARGET)
+    </k>
+    requires isIntrinsicFunction(FUNC)
+     andBool notBool #functionNameMatchesEnv(getFunctionName(FUNC))
```
mariaKt: Could you clarify when the fall backs wold be used? The intercept already matches on `#isSPLSolMemsetFunc`, so `FUNC` is `sol_memset`. When would `isIntrinsicFunction(FUNC)` be true for `sol_memset`? If never, the fallbacks for intrinsic are not needed.

mariaKt (approved):
Approved, with minor comments

kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
Stevengre: Good catch. I removed the intrinsic fallbacks in 3f2d282c. This hook is entered only after `#isSPLSolMemsetFunc(#functionName(FUNC))`, so `FUNC` is the ordinary `solana_program_memory::sol_memset` mono item, not an `IntrinsicFunction(_)`. The remaining fallbacks are the reachable ordinary-call paths.

kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
```diff
@@ -602,6 +673,87 @@ The `#initBorrow` helper resets borrow counters to 0 and sets the correct dynami
...
+          DEST, TARGET, UNWIND, SPAN) ~> _CONT
+      => #execSPLSolMemset(FTY, FUNC, #withDeref(BUF), VAL, LEN, BUF, BUF VAL LEN .Operands, DEST, TARGET, UNWIND, SPAN)
+    </k>
+    requires #isSPLSolMemsetFunc(#functionName(FUNC))
+    [priority(30), preserves-definedness]
+
+  syntax KItem ::= #execSPLSolMemset ( Ty, MonoItemKind, Evaluation , Evaluation , Evaluation , Operand, Operands, Place, MaybeBasicBlockIdx, UnwindAction, Span ) [seqstrict(3,4,5)]
+
+  rule <k> #execSPLSolMemset(_, _, SPLDataBuffer(_) #as BUF, VAL, Integer(LEN, 64, false), operandCopy(DESTBUF), _ARGS, _DEST, TARGET, _UNWIND, _SPAN)
+        => #setLocalValue(DESTBUF, SPLDataBuffer(Integer(0, 8, false))) ~> #continueAt(TARGET) ... </k>
```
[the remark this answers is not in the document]
Stevengre: Fixed in 0310b4b2. The fast-path now writes through the dereferenced buffer place, matching the dispatchers `#withDeref(BUF)` read path and the existing pointer-write intrinsic pattern. That preserves the pointer local and updates the pointee `SPLDataBuffer` instead of corrupting the local pointer value.

kmir/src/kmir/kdist/mir-semantics/symbolic/spl-token.md
Stevengre: Added a clarifying comment in 3f2d282c. This `99` is the multisig buffer size: 1 byte for `m`, 1 for `n`, 1 for `is_initialized`, plus `MAX_SIGNERS * 32` signer bytes. In this SPL symbolic model `MAX_SIGNERS = 3`, so the total is `3 + 3 * 32 = 99`.

PR #1009, opened 2026-03-31
test: enable --terminate-on-thunk for all prove tests

Stevengre:
## Summary
- Make `--terminate-on-thunk` the default for all prove tests instead of per-test opt-in via `PROVE_TERMINATE_ON_THUNK`
- This exposes unresolved thunks (from incomplete semantics) that were previously silently propagated through proof execution
- All 110 prove tests pass with this change on master — no regressions
- Update 13 show expected files whose proof trees changed due to earlier thunk termination

## Context
First introduced in PR #813, `--terminate-on-thunk` makes the prover treat any `thunk(...)` in the K cell as a terminal state, immediately stopping proof exploration. Previously only `closure-staged` opted in. Enabling it globally ensures we catch any test that silently relies on thunk propagation.

## Test plan
- [x] All 110 `test_prove` parametrizations pass locally
- [x] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dkcumming (approved):
Awesome! Let's give this a go, and if it works out we should consider removing `thunk` alltoghter

PR #1010, opened 2026-04-01
feat(symbolic): add spl-token lemma for inner_test_validate_owner

Stevengre:
## Summary

SPL-Token version of the `inner_test_validate_owner` lemma rules (adapted from the p-token version in PR #1001).

## Key differences from p-token lemma

- **Different intercept strategy**: spl-token's `AccountInfo` uses raw `Aggregate` values instead of p-token's `PAccountAccount`/`PAcc` wrappers. The intercept rules extract individual fields (`key`, `is_signer`, `owner`) from `AccountInfo` using field projections (fields 0, 5, 3 respectively).
- **Simplified case rules**: match pre-extracted `Value`s directly instead of destructuring `PAccount` wrappers.
- **Symbolic bool handling**: uses `_IS_SIGNER` wildcard since spl-token's `is_signer` may be a symbolic `BoolVal(VAR:Bool)` that doesn't match `BoolVal(true)` or `BoolVal(false)`.
- **Case 1b**: handles keys-match + `Err(Custom(4))` path via LHS pattern match (avoids booster sort injection errors from `requires ==K`).
- **Cases 8-10**: multisig signer-checking falls through to small-step execution (sufficient for n==1).

## Also included

- Fix `#forceSetLocal` → `#setLocalValue` in `spl-token.md` for compatibility with `origin/master` (which removed the redundant helper in PR #810).

## Proof result

`test_validate_owner_multisig` (spl-token, n==1):
- **With lemma**: ✅ PASSED, 27 nodes, 2886s
- **Without lemma**: ✅ PASSED, 123 nodes, 4877s
- **Speedup**: 41% faster, 78% fewer nodes

## Test plan

- [x] `test_validate_owner_multisig` proof passes with lemma enabled (n==1 constraint)
- [x] K semantics builds successfully (`make build`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,12 +47,37 @@ jobs:
         with:
           version: ${{ steps.uv_release.outputs.uv_version }}
       - name: 'Run unit tests'
-        run: make test-unit
+        run: make test-unit PARALLEL=12
 
   integration-tests:
     needs: code-quality-checks
-    name: 'Integration Tests'
+    name: 'Integration Tests ${{ matrix.name }}'
     runs-on: [self-hosted, linux, normal]
+    timeout-minutes: ${{ matrix.timeout }}
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - name: 'LLVM Concrete Tests'
+            test-args: '-k "llvm or test_run_smir_random"'
+            parallel: 12
+            timeout: 20
+          - name: 'Haskell Exec SMIR'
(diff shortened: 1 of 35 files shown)


Stevengre:
Closing this out as well. This PR is the SPL-token counterpart to #1001 and follows the same validate_owner lemma approach. Since we are no longer pursuing that direction, this PR is no longer needed. If any SPL-specific piece becomes useful later, we can bring it back in a smaller fresh PR.

PR #1007, opened 2026-03-31
chore: sync feature/p-token with master

Stevengre:
## Summary
- sync `feature/p-token` with current `master` so the symbolic SPL work carries the latest runtime, type, test, and CI updates
- resolve the stale `#forceSetLocal` usage exposed by the merge with current master
- refresh the imported prove-show expected outputs so they match the merged branch current rendering

## Upstream commits included
- `Teach kmir to reduce SMIR K definitions over a set of CFG roots` (#845)
- `ci: increase integration test timeouts` (#1008)
- `fix(rt): handle zero-sized function types in decode and ZST recognition` (#813)
- `fix(types): make #pointeeProjection confluent with source-first unwrapping strategy` (#988)
- `cleanup(rt): remove redundant #forceSetLocal` (#810)
- `Basic support for floats` (#995)
- `perf(test): remove redundant integration test executions` (#972)
- ``volatile_{load,store}`` with statics, heap-allocation repros, and transparent-wrapper transmute coverage (#989)

## Why a replacement PR
`feature/p-token` is a protected branch, so the original `master -> feature/p-token` sync PR cannot be repaired by directly updating the base branch. This replacement PR carries the resolved merge result on `sync-pr1006` instead.

Supersedes #1006.

## Validation
- `make build`
- `make check`
- `make test-unit PARALLEL=12`
- `make smir-parse-tests`
- refreshed the six imported `prove-rs/show` expectations via scoped `--update-expected-output`
- GitHub Actions on this replacement PR are the source of truth for the full post-merge matrix

PR #1012, opened 2026-04-02
ci: install nix on MacM1 runners

Stevengre:
## Summary
- run the existing Nix setup steps on `MacM1` runners instead of skipping them
- keep the `nix` job logic identical across `normal` and `MacM1` so `nix --version` and `nix flake check` run against an initialized environment

## Why
The current workflow skips `Install Nix` and `Install Cachix` on `MacM1`, but the following `Build and test` step still runs `nix --version` and `nix flake check`. That makes the MacM1 job fail before it can exercise the actual repository checks.

## Validation
- workflow diff reviewed against `master`
- no repository code paths changed; only the Nix job setup conditions were removed

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -220,7 +220,6 @@ jobs:
           submodules: recursive
 
       - name: 'Install Nix'
-        if: ${{ matrix.runner != 'MacM1' }}
         uses: cachix/install-nix-action@v31.5.1
         with:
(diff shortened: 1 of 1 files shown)


Stevengre:
> PR https://github.com/runtimeverification/mir-semantics/pull/1007 is the failing PR that prompted this one?

yes, I restart several times, but it doesn't work.

PR #1014, opened 2026-04-03
fix(pointer-casts): support array-to-wrapper projections

Stevengre:
## Summary
- relax transparent-wrapper pointee projection so array-to-wrapper pointer casts can advance
- cover wrapper, nested-wrapper, and singleton-wrapped-array pointer cast regressions with one shared semantic fix

## Test plan
- `cd kmir && uv run pytest src/tests/integration/test_integration.py -k "ptr-cast-array-to-wrapper-fail or ptr-cast-array-to-nested-wrapper-fail or ptr-cast-array-to-singleton-wrapped-array-fail or ptr-through-wrapper-fail or raw-ptr-cast-fail or ref-ptr-cast-elem-fail or ref-ptr-cast-elem-offset-fail or pointer-cast-length-test-fail" -q`


Stevengre:
Closing: the array-to-wrapper pointer cast regressions this PR targeted are addressed by #1003 (stable address allocation model). The semantic rule changes here had no observable effect on the claimed test cases — all 8 proofs stuck at the same `thunk(#cast(...))` point with identical proof trees before and after the changes.

PR #1015, opened 2026-04-03
refactor(tests): distinguish -fail from -unsupported, auto-detect show specs

Stevengre:
## Summary
- Rename 20 prove-rs test files from `-fail` to `-unsupported` to distinguish tests that fail because the semantics doesn't support a feature yet from tests where the program itself is expected to fail
- Remove manually maintained `PROVE_SHOW_SPECS` list — all `-fail` and `-unsupported` tests now automatically verify show output
- Remove `assert!(false)` from `symbolic-args` test (main passes with concrete args, only `eats_all_args` is unsupported)
- Switch CLI test fixtures (`test_cli_show_statistics_and_leaves`, `test_cli_show_minimize_proof`) to use `symbolic-structs-fail.eats_struct_args` for stable proof trees with symbolic branching

## Test plan
- `cd kmir && uv run pytest src/tests/integration/test_integration.py -k "unsupported or fail" -q`
- `cd kmir && uv run pytest src/tests/integration/test_cli.py -k "statistics_and_leaves or minimize_proof" -q`


kmir/src/tests/integration/test_integration.py
```diff
@@ -99,7 +65,8 @@ def test_prove(rs_file: Path, kmir: KMIR, update_expected_output: bool) -> None:
         prove_opts.start_symbol = start_symbol
         apr_proof = kmir.prove_program(prove_opts)
 
-        if should_show:
+        if should_fail:
+            assert apr_proof.failed
```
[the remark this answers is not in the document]
Stevengre: Fixed in f8f58dcc. `test_prove` now only compares show snapshots when the expected file already exists during normal runs, while `--update-expected-output` still creates missing fixtures. I also re-ran the reviewer-named cases: `assert-false-fail` and `loop-unsupported` no longer fail on missing snapshot files; `enum-two-refs-unsupported` still fails earlier for an unrelated `.smir.json` generation issue.

.github/workflows/test.yml
```diff
@@ -77,7 +77,7 @@ jobs:
           - name: 'Verify Rust Std'
             test-args: '-k test_verify_rust_std'
             parallel: 6
-            timeout: 60
+            timeout: 90
```
dkcumming, line 80: I think this should be a redundant change since #1069 improved performance, although I haven't seen a repeat on runner 10:
- https://github.com/runtimeverification/mir-semantics/actions/runs/24694894839/job/72225410402?pr=1069
- https://github.com/runtimeverification/mir-semantics/actions/runs/24697098086/job/72232173183?pr=1071

Not necessarily requesting a change just noting that it _should_ be less than 60 consistently now.

kmir/src/kmir/testing/fixtures.py
```diff
@@ -31,20 +31,26 @@ def _normalize_symbol_hashes(text: str) -> str:
     return text
 
 
+def _normalize_show_output(text: str) -> str:
+    text = _normalize_symbol_hashes(text)
+    text = re.sub(r'(?m)^(\s*(?:[│┃┊]\s*)?span: )\d+$', r'\1<span>', text)
+    text = re.sub(r'(?m)^\s*>> message: .*\n?', '', text)
+    return text.rstrip('\r\n')
```
dkcumming, line 38: Is there a reason we want to remove the decoded message? If we have a different match with a fail that  errors with a message and a span, would it not be helpful to have that information? Is having the information creating a problem?

I am always deeply suspicious of claude using regex covering things up. At the very least this should be adequately documented to explain what it is doing.

kmir/src/kmir/cargo.py
```diff
@@ -182,6 +182,9 @@ def cargo_get_smir_json(
     cwd = cwd or Path.cwd()
     smir_json_result = cwd / rs_file.with_suffix('.smir.json').name
     run_process_2(command, cwd=cwd)
+    resolved_smir_json_result = cwd / rs_file.resolve().with_suffix('.smir.json').name
+    if not smir_json_result.is_file() and resolved_smir_json_result.is_file():
+        resolved_smir_json_result.rename(smir_json_result)
```
dkcumming, line 187: I couldn't understand the point of this until I queried claude. It might be nice to mention that this is addressing symlinks.

dkcumming (approved):
I like this! I have the comments that would be nice to consider and/or address, but I don't want to get in the way of this merging so approving!

PR #1018, opened 2026-04-07
Decompose multisig signer pubkeys for p-token

mariaKt:
Similar to https://github.com/runtimeverification/mir-semantics/pull/1000, but for p-token.

I noticed the new tests for 3 signers failing to prune paths with contradictory conditions: a specific `tx_signer` `key` already has a byte-by-byte key mismatch with a registered signer key, but then due to the signer being represented as a `List` , the same comparison is re-evaluated using `?Signer0:List` (opaque) against the same `tx_signer` bytes, with `==K` being opaque.

### DIFF
--- a/kmir/src/kmir/kdist/mir-semantics/symbolic/p-token.md
+++ b/kmir/src/kmir/kdist/mir-semantics/symbolic/p-token.md
@@ -110,6 +110,39 @@ The code uses some helper sorts for better readability.
   rule allBytes( ListItem(Integer(_, 8, false)) REST:List) => allBytes(REST)
   rule allBytes( ListItem(_OTHER) _:List) => false [owise]
 
+  // Construct a 32-byte pubkey List from individual Int variables.
+  // When used with existential variables (?Var:Int), this produces a concrete List structure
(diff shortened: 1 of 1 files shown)


mariaKt:
I have tested this with a few p-token proofs, though not all (my machine cannot run many proofs in parallel). The proofs with 1 `tx_signer` pass:
- `approve`, `withdraw_excess_lamports_multisig`: these do not exercise the changes, just checking a couple to be safe
- `initialize_multisig`, `initialize_multisig2`: these use `cheatcode_is_multisig`
- `approve_multisig`, `approve_checked_multisig`, `withdraw_excess_lamports_multisig_multisig`, `revoke_multisig`: these use 'cheatcode_is_multisig' and test the decomposed pubkeys through key comparisons in `validate_owner`
 I am waiting on the results of one the new tests with 3 `tx_signer` (using the additional fix [here](https://github.com/runtimeverification/solana-token/pull/190) locally). I will run more after the 3 signer proof is done.

Stevengre (approved):
LGTM!

PR #1023, opened 2026-04-07
Another increase of timeout for Haskell Exec SMIR CI test

mariaKt:
The timeout increase [here](https://github.com/runtimeverification/mir-semantics/pull/1021) is not enough (depending on the state of the GH runners, this is still a bit too tight). I measured locally, and it took around 40 minutes, so I added another 10 as a leeway for a total of 50 min.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,7 +65,7 @@ jobs:
           - name: 'Haskell Exec SMIR'
             test-args: '-k "test_exec_smir and haskell"'
             parallel: 6
-            timeout: 30
+            timeout: 50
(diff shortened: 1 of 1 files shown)


jberthold (commented):
LGTM. ⏱️ I'd even go with 60 minutes here... github runners may be noisy/weak and the timeout is just an emergency break.

mariaKt:
> LGTM. ⏱️ I'd even go with 60 minutes here... github runners may be noisy/weak and the timeout is just an emergency break.

Got it, will do. Thank you.

palinatolmach (approved):

PR #1056, opened 2026-04-10
refactor(rt): use frame ids in stack references

Stevengre:
## Summary
This draft refactors stack-based `Reference`/`PtrLocal` values to use stable frame ids instead of relative stack offsets.

## What changed
- keep `StackFrame` frame ids as stable runtime identities
- make stack references and raw local pointers carry frame ids directly
- remove offset-adjustment helper layer and the corresponding lemmas
- resolve stack dereferences and writes by frame id
- update branch-specific expected outputs affected by the new representation

## Validation
- `make build`
- `uv --directory kmir run pytest src/tests/unit -q`
- `uv --directory kmir run pytest src/tests/integration/test_decode_value.py -q`
- `uv --directory kmir run pytest src/tests/integration/test_integration.py -q -x -k 'Ref-weirdRefs'`
- `uv --directory kmir run pytest src/tests/integration/test_integration.py -q -x -k 'main-a-b-c or closure-call'`
- `uv --directory kmir run pytest src/tests/integration/test_integration.py -q -x -k 'Ref-refAsArg or Ref-refReturned or Ref-doubleRef or iterator-simple'`
- `uv --directory kmir run pytest src/tests/integration/test_run_smir_random.py -q`
- `uv --directory kmir run pytest src/tests/integration/test_cli.py -q -k 'show_to_module or show_statistics_and_leaves or show_printers_snapshot'`

## Follow-up
PR #1003 can now be based on this draft and keep its pointer-cast/address-allocation work separate from the stack-reference-model refactor.


Stevengre:
We don't need this because we can use `size(stack) - OFFSET` to get a stable identifier for frame.

Issue #1060, opened 2026-04-14
Improvement to `kmir show`


mariaKt:
`kmir show` holds entire output in memory, resulting in large memory consumption and causing OOM on large proofs ([slack](https://runtimeverification.slack.com/archives/C04LQ1L84TG/p1776183737835549?thread_ts=1776176595.005889&cid=C04LQ1L84TG)).

`kmir show` (both `--full-printer` and `--statistics --leaves`) builds the complete output as a list of strings in memory, then joins them into a single string before printing ([`__main__.py`, lines 167-191](https://github.com/runtimeverification/mir-semantics/blob/9ada694243298fcbdcad784ea2d9442f57dd6840/kmir/src/kmir/__main__.py#L164)). For large proofs (e.g., 2000+ nodes), this causes out-of-memory failures.

The issue is in `_kmir_show()`:
1. `shower.show()` returns a full `list[str]` of all lines
2. `render_statistics()` / `render_leaf_k_cells()` extend the list further
3. `'\n'.join(lines)` creates a second copy as one string
4. The joined string is printed to stdout

The output is effectively held in memory twice (list + joined string).

**Proposed fix:** Stream lines to stdout as they are generated instead of accumulating them. Convert `shower.show()`, `render_statistics()`, and `render_leaf_k_cells()` to generators (or accept a write callback), and print/flush each line immediately.

**Observed impact:** `kmir show --full-printer` OOM'd after a successful 60h proof of `transfer_multisig` ([here](https://kaas.runtimeverification.com/app/organization/runtimeverification/solana-token/job/5F33ASWNe3Tf4B2qjEcq3E?job_hl=vH7kuxXApyrUUpUE7NKc35)), producing a 0-byte output file. The proof result (`proof.json`, `kcfg/`) was intact but the stats could not be generated.

mariaKt:
Closed with https://github.com/runtimeverification/mir-semantics/pull/1062.

PR #1062, opened 2026-04-14
`kmir show` streams output instead of joining into one string

mariaKt:
This PR addresses https://github.com/runtimeverification/mir-semantics/issues/1060.

`kmir show` streams output to reduce memory usage

`kmir show` previously accumulated the entire output as a list of strings, then joined them into a single string before printing to stdout. For large proofs (2000+ nodes), this caused out-of-memory failures during stats generation — even after the proof itself completed successfully.

This PR changes `_kmir_show()` to stream each section's output to stdout independently, freeing memory between sections:

1. The main tree output (`shower.show()`) is printed line-by-line and released before statistics are generated
2. `render_statistics()`, `render_rules()`, and `render_leaf_k_cells()` are each printed and released independently
3. Blank line separators between sections are added conditionally (matching original behavior) to avoid double-blank lines

A helper `_flush_lines()` accepts accepting any iterable for flexibility. Note that the remaining peak memory cost (`shower.show()` building the full tree output in memory) is on pyk's side (`APRProofShow.show()` returns `list[str]`).

**Observed impact:** `kmir show` OOM'd after a large successful proof ([transfer_multisig](https://kaas.runtimeverification.com/app/organization/runtimeverification/solana-token/job/5F33ASWNe3Tf4B2qjEcq3E?job_hl=vH7kuxXApyrUUpUE7NKc35)). The proof data was intact but stats could not be generated. With this change, each section's memory is freed before the next one starts.

**Testing:** Output is identical to the previous version, verified by diff against existing proof outputs (both standalone and when appending a header via `run-proofs.sh`), and by the `test_cli_show_*` integration tests.

### DIFF
--- a/kmir/src/kmir/__main__.py
+++ b/kmir/src/kmir/__main__.py
@@ -193,16 +193,23 @@ def _kmir_show(opts: ShowOpts) -> None:
         print(f'Module written to: {opts.to_module}')
     else:
         # Stream lines to stdout to avoid holding the entire output in memory
+        last_line = lines[-1] if lines else ''
         _flush_lines(lines)
         del lines
         if opts.statistics:
-            print()
-            _flush_lines(render_statistics(proof))
+            if last_line != '':
+                print()
+            stat_lines = render_statistics(proof)
+            last_line = stat_lines[-1] if stat_lines else last_line
+            _flush_lines(stat_lines)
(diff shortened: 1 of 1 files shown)


Stevengre (approved):
LGTM!

PR #1069, opened 2026-04-17
Reuse prebuilt KMIR definition to avoid redundant `llvm-kompile` processes

dkcumming:
If a Rust file contains multiple targets to prove (e.g. [verify-rust-std challenge 11](https://github.com/runtimeverification/mir-semantics/tree/master/kmir/src/tests/integration/data/verify-rust-std/0011-floats-ints)) then multiple `llvm-kompile` processes will be invoked by [_prove](https://github.com/runtimeverification/mir-semantics/blob/c92c8ce0b4b0d762d679a15351b562fa243fd8f0/kmir/src/kmir/_prove.py#L70-L77). This PR pushes this process earlier to create a KMIR definition with all the required start symbols so that `llvm-kompile` is called once per file.

In particular:
- `test_verify_rust_std` now compiles a prebuilt KMIR definition with reduced start symbols. This definition is reused for all proof harnesses in a file;
- `prove_with_kmir` is added as separate control-flow to `prove`. This may not be _elegant_, but right now I think it is the clearest separation, and I was hesitant to add flags or refactor too aggressively;

Results from verify-std-rust challenge 11:
- Prior to this refactoring: 1 `llvm-kompile` / start-symbol = 138 `llvm-kompile` invocations
- After this refactoring: 1 `llvm-kompile` / file = 16 `llvm-kompile` invocations
- Approximately 3x increase of performance

<details>

PRIOR:

real  31m16.872s
user 170m18.245s
 sys   5m51.240s

real  31m36.755s
user 175m15.830s
 sys   6m32.859s

AFTER:

real    10m20.730s
user    44m10.271s
sys     2m15.364s

real 10m58s
user 47m0s
 sys  2m24s

real 10m40.395s
user 45m54.822s
 sys  2m20.399s
 
real 10m38.370s
user 45m32.886s
 sys  2m19.782s

</details>


kmir/src/kmir/_prove.py
```diff
@@ -49,6 +50,33 @@ def prove(opts: ProveOpts) -> APRProof:
         return _prove(opts, target_path, label)
 
 
+def prove_with_kmir(
+    kmir: KMIR,
+    smir_info: SMIRInfo,
+    opts: ProveOpts,
+) -> APRProof:
```
Stevengre, line 57: Thank you! I think it's really a good idea. But I see some redundant lines of code here compared to `_prove`. What do you think to make `kmir` as a `ProveOpts` so that we can separate the proof prepare and proof run process?

dkcumming: There is some duplication, and I did consider merging them. In the end `prove` is changed in the stacked PR #1071 . I would be open to discussing the design and improvements in the context of that PR.

mariaKt (approved):
LGTM, especially since there is further refactoring to address the duplication issues.

kmir/src/kmir/_prove.py
Stevengre: GREAT!

PR #1071, opened 2026-04-20
`kmir prove` takes multiple `start-symbols`

dkcumming:
This PR adds the ability to provide multiple start symbols to `kmir prove`, and importantly, a single call to `llvm-kompile` for the proofs.

In particular:
- Names can be provided to `--start-symbol(s)` either comma separated or repeated `kmir prove --start-symbols name1, name2 --start-symbol name3`;
- Function `_prove_multi` will reduce the items to the provided names and do a single `llvm-kompile` to be shared;
- Each proof is run _sequentially_ and the definition and smir.json is stored in a directory `filename_stem.kompiled`, and the proof dirs just store the kcfg (this is only if multiple start symbols are provided);
- `prove_with_kmir` is not removed. It was considered but the way it interacts with the test runner is nicer so I left it;
- Test is added to `test_cli`;
- Observed to correctly reduce the calls to `llvm-kompile` and has greatly increased performance for mutliple start symbol proofs;


kmir/src/kmir/_prove.py
```diff
@@ -31,23 +31,26 @@
...
+            target_path = opts.proof_dir / f'{opts.rs_file.stem}.kompiled'
+        return _prove_multi(opts, target_path)
 
     with tempfile.TemporaryDirectory() as tmp_dir:
         target_path = Path(tmp_dir)
-        return _prove(opts, target_path, label)
+        return _prove_multi(opts, target_path)
 
 
 def prove_with_kmir(
```
mariaKt, line 56: Since you prefer to keep `prove_with_kmir`, please comment/document when its use makes sense / when it would be preferred (prove_with_kmir vs? prove). It uses opts.start_symbols[0] (which is fine for the test runner but could silently reject additional symbols). Feel free to address as seems appropriate to you (assert/document the intended uses, ?).

mariaKt (approved):
Approved, with one comment.

kmir/src/kmir/_prove.py
dkcumming: Oh that is true. I will add documentation and asserts now

kmir/src/kmir/__main__.py
```diff
@@ -587,7 +598,14 @@ def _arg_parser() -> ArgumentParser:
...
     prove_parser.add_argument(
-        '--start-symbol', type=str, metavar='SYMBOL', default='main', help='Symbol name to begin execution from'
+        '--start-symbol',
+        '--start-symbols',
+        dest='start_symbols',
+        type=str,
+        metavar='SYMBOL',
+        action='append',
+        default=None,
+        help='Symbol name(s) to prove (repeatable, comma-separated allowed)',
```
ehildenb, line 608:
```suggestion
        '--start-symbol',
        '--start-symbols',
        dest='start_symbols',
        type=list,
        metavar='SYMBOL',
        action='append',
        default=None,
        help='Symbol name(s) to prove (repeatable, comma-separated allowed)',
```

I think with `action='append'` it should be `type=list`? I may be mistaken there.

PR #1072, opened 2026-04-21
refactor(kmir): simplify terminator call setup

Stevengre:
## Summary
- Refactor terminator call execution into separate target preparation, caller-frame operand materialization, and callee-frame execution steps.
- Match body-bearing `monoItemFn` calls directly and keep no-body calls stuck at `#prepareTerminatorCall` for annotation.
- Mark the `kmir.md` helper functions as total, adding fallback behavior where the return sort can represent the fallback.
- Update no-body function annotation in `kmir/src/kmir/utils.py` to recognize the new stuck call shape.

## Testing
- `git diff --check -- kmir/src/kmir/kdist/mir-semantics/kmir.md kmir/src/kmir/utils.py`
- `make build`
- `make check`
- `make test-integration TEST_ARGS="--update-expected-output"` was started and then interrupted at about 17%; generated expected-output updates remain unstaged and are not included in this draft PR.

## Known Limitations / Follow-ups
- Expected-output updates are still local unstaged changes and need a separate follow-up commit before this draft is ready for full review.


kmir/src/kmir/kdist/mir-semantics/kmir.md
```diff
@@ -333,59 +337,100 @@ where the returned result should go.
...
+
+  rule <k> #prepareBodyCall(ACC, NAME, BODY, ORIGINAL, OP:Operand REST:Operands, CALLERLOCALS, FTY, DEST, TARGET, UNWIND, SPAN)
+        => OP ~> #prepareBodyCall(ACC, NAME, BODY, ORIGINAL, REST, CALLERLOCALS, FTY, DEST, TARGET, UNWIND, SPAN)
+        ...
+       </k>
+
+  rule <k> VAL:Value ~> #prepareBodyCall(ACC, NAME, BODY, ORIGINAL, REST:Operands, CALLERLOCALS, FTY, DEST, TARGET, UNWIND, SPAN)
+        => #prepareBodyCall(ACC ListItem(VAL), NAME, BODY, ORIGINAL, REST, CALLERLOCALS, FTY, DEST, TARGET, UNWIND, SPAN)
+        ...
+       </k>
```
ehildenb, line 367: This looks like `#prepareBodyCall` is basically used to heat/cool the `Operands` into `Value`, is that right? In that case, could we just mark `#execTerminatorCall` as `strict` on the 5th argument, and have heating and cooling take care of it? It could heat the entire list of `Operands`, turn them into `Values`, then cool it back in? Not sure if it would work as currently configured, but maybe it would be a simpler approach?

Stevengre: Fixed here: f92b40d7879c298c65379673d11e1af5f2d0d63e

I use `#readOperands` to achieve this goal, but the side effect is to add `syntax KResult := List`. Another way to solve this might change `#readOperands` implementation and generate a `Range` Value in the end.

What's your idea?

PR #1079, opened 2026-04-27
Avoid llvm-kompile when running proofs - rebooted

jberthold:
* Do not extend and recompile the definition.kore file for LLVM, only extend the Haskell one
* Mark the `lookupTy`, `lookupAlloc` and `lookupFunction` as `no-evaluators` (including the stratified helper functions)
* Also mark (transitively) all functions whose equations contain these lookup functions or other `no-evaluators` functions on the RHS as `no-evaluators` (the LLVM backend will core-dump when it encounters a `no-evaluators` function).

Integration tests run considerably faster. 
- [x] Integration tests
- [x] Test this with Solana token proofs


dkcumming (approved):
Awesome stuff. Mir semantics test suite passes in 30% of time previously. Solana P-Token proofs also appear to have slight improvement.

Stevengre:
Amazing!

PR #1081, opened 2026-05-05
Support for .rs files in kmir run

mariaKt:
This PR changes the interface of `kmir run` to match `prove`, and adds to it the ability to accept `.rs` files directly.
- Positional FILE argument (mandatory): Rust file by default, SMIR JSON if `--smir`, cargo target if `--bin`
- `--smir` and `--bin` are mutually exclusive
- `--save-smir` option to keep the intermediate SMIR JSON when running from .rs
- Updated documentation in README and exec-smir README

Finally, I renamed the `test_stable_mir_ui_pass.py` (which uses `prove`) to `test_stable_mir_ui_prove.py`, and provided the `test_stable_mir_ui_exec.py` that uses `run`.


kmir/src/kmir/__main__.py
```diff
@@ -657,11 +659,13 @@ def _parse_args(ns: Namespace) -> KMirOpts:
     match ns.command:
         case 'run':
             return RunOpts(
+                rs_file=ns.rs_file,
+                start_symbol=ns.start_symbol,
```
dkcumming, line 663: Originally, I was thinking a run would always invoke a Rust program from the default entry point (typically `main`), but I tested this on [sum-to-n-functions.rs](https://github.com/runtimeverification/mir-semantics/blob/mk/kmir-run-rs/kmir/src/tests/integration/data/prove-rs/sum-to-n-functions.rs) and it works with start symbol of either "main" or "test_sum_to_n" which is actually quite nice. I wasn't sure what would happen if it was pointed to "sum_to_n" (since that takes arguments which we would have to somehow instantiate) but it failed with a nice error 
```
ValueError: Cannot create concrete call configuration for sum_to_n: function has parameters
```

dkcumming (approved):
Nice all looks great to me. I tested it and I can't find any flaws!

PR #1083, opened 2026-05-08
Implementation of easy intrinsics and skip lists updates

mariaKt:
This PR implements a few intrinsics that we identified missing from running the ui test suite.

### Skip lists

- Sorted the skip list for consistent ordering
- Separated skip lists per backend (renamed `skip.txt` -> `skip-prove.txt`, added `skip-exec.txt`), since the two backends can have different pass/fail sets.
- Updated skip lists based on full runs of all 2,888 stable-mir-ui tests:
  - **Haskell backend** (`skip-prove.txt`): 1,159 pass, 1,729 fail
  - **LLVM backend** (`skip-exec.txt`): 1,164 pass, 1,724 fail
- Added `test-stable-mir-ui-exec` Makefile target for LLVM UI tests (renamed existing target to `test-stable-mir-ui-prove`)
- Enabled `--update-skip` for the LLVM test suite (`test_stable_mir_ui_exec.py`)
- Removed 2 tests that now pass on both backends after the intrinsic fixes below

### New intrinsics

Each intrinsic is preceded by a test commit (with expected output showing the proof stuck on the missing intrinsic) and followed by an updated expected output showing the proof passing.

- `rotate_left` — bitwise left rotation using `#rotateLeftInt` helper (2 previously failing tests)
- `bswap` — byte swap using `Int2Bytes`/`Bytes2Int` endianness conversion (1 previously failing test)
- `ctpop` — population count using recursive `#popcount` helper (3 previously failing tests)
- `ctlz_nonzero` — count leading zeros using `#log2floor` helper (1 previously failing test)

Of the 7 tests previously stuck on these intrinsics, 2 now pass end-to-end on both backends:
- `rfcs/rfc-2294-if-let-guard/run-pass.rs` (was stuck on `rotate_left`)
- `consts/const-endianess.rs` (was stuck on `bswap`)

The other 5 progress past the intrinsic but hit further issues.


kmir/src/kmir/kdist/mir-semantics/intrinsics.md
```diff
@@ -182,6 +182,99 @@ the dereferenced operand is evaluated (i.e., the value is read from memory) befo
...
+       ... </k>
+
+  syntax Int ::= #log2floor(Int) [function]
+  rule #log2floor(1) => 0
+  rule #log2floor(N) => 1 +Int #log2floor(N >>Int 1)
+    requires N >Int 1
+
+  rule <k> #execCtlz(DEST, Integer(VAL, WIDTH, SIGN))
+        => #setLocalValue(DEST, Integer(WIDTH -Int 1 -Int #log2floor(truncate(VAL, WIDTH, Unsigned)), WIDTH, SIGN))
+       ... </k>
```
dkcumming: I think this is equivalent to using [log2Int from domains.md](https://github.com/runtimeverification/k/blob/6ee29000016c6addb83b39ac865860f9542b3ebf/k-distribution/include/kframework/builtin/domains.md?plain=1#L1296-L1305) 

```suggestion
  rule <k> #execCtlz(DEST, Integer(VAL, WIDTH, SIGN))
        => #setLocalValue(DEST, Integer(WIDTH -Int 1 -Int log2Int(truncate(VAL, WIDTH, Unsigned)), WIDTH, SIGN))
       ... </k>
```

dkcumming (approved):
Nice! I added some clean up of the test files to a commit on the end. I also think that we could use `log2Int`, I think this has the same semantics but maybe it is different for zero input which "is `#False`". Does that mean that the branch would get stuck?

kmir/src/kmir/kdist/mir-semantics/intrinsics.md
mariaKt: We implement cltz_nonzero (which expects non zero input, 0 is UB), and the requires makes sure of that. So yes, it would get stuck, but we want it to. The replacement makes sense, I committed this.

PR #1092, opened 2026-06-26
`size_of_val`, `min_align_of_val` implementation

mariaKt:
This PR implements `size_of_val`, `min_align_of_val` for sized types.

I show the stuck state and passing state before and after each one's implementation, and remove the output files at the end (included only to help review).

We do not handle unsized types atm, left for future work (their dynamic size calculation is likely to change).

Finally, we update the skip lists for the LLVM and Haskell backend. Note that these newly passing tests reflect not just this PR, but also PR [#1091](https://github.com/runtimeverification/mir-semantics/pull/1091).


kmir/src/kmir/kdist/mir-semantics/intrinsics.md
```diff
@@ -319,6 +319,46 @@ the second argument, so the returned difference is always positive.
...
+
+```k
+  rule <k> #execIntrinsic(IntrinsicFunction(symbol("size_of_val")), ARG:Operand .Operands, DEST, _SPAN)
+        => #setLocalValue(DEST, Integer(#sizeOf(TYPESMAP, lookupTy(TYPESMAP, {pointeeTy(lookupTy(TYPESMAP, {#extractOperandType(TYPESMAP, ARG, LOCALS)}:>Ty))}:>Ty)), 64, false))
+       ... </k>
+       <locals> LOCALS </locals>
+       <types> TYPESMAP </types>
+    requires isTy(#extractOperandType(TYPESMAP, ARG, LOCALS))
+     andBool isTy(pointeeTy(lookupTy(TYPESMAP, {#extractOperandType(TYPESMAP, ARG, LOCALS)}:>Ty)))
+     andBool #metadataSize(TYPESMAP, pointeeTy(lookupTy(TYPESMAP, {#extractOperandType(TYPESMAP, ARG, LOCALS)}:>Ty))) =/=K dynamicSize(1)
```
dkcumming, line 340: At first I thought this was too permissive as it only restricts `dynamicSize(1)` for all slices. However, looking at [rt/types.md](https://github.com/runtimeverification/mir-semantics/blob/4d793252bcd77091ee759ca6cd1629db41ed5496/kmir/src/kmir/kdist/mir-semantics/rt/types.md?plain=1#L416-L441) the codomain of `#metadataSize` is `{ staticSize(N), dynamicSize(1), noMetadataSize }`.  So in this case, `dynamicSize(1)` is a sentinel value only.

Not sure if it is worth changing but it would be nice to continue the pattern of using sentinels that are logically distinct from sensible values. I think `dynamicSize(-1)` would be better here because if one knows that it is impossible immediately by reading it so either there is an error or it is a sentinel only.

dkcumming (approved):
Nice! I left a comment about the sentinel value that I think we should consider. But approving as it should not block this.

